PR #1, opened 2019-06-06
Fix custom install to work

Vash63:
Custom install location breaks for all default Steam installs as the space in the directory 'Quake 2' causes only the first part of the path to be passed. Fixes error: ```./find-retail-paks.sh: line 14: pushd: /mnt/steam/steamapps/common/Quake: No such file or directory```


apanteleev:
Thank you for the contribution!

Vash63:
Great, thanks for the fast merge! Hope to see it on Steam soon for new users.

PR #3, opened 2019-06-06
Fixing CMAKE Version in README

namar0x0309:


apanteleev:
Why do you suggest to increase the min cmake version?

namar0x0309:
@apanteleev On here https://cmake.org/download/, the highest version is 3.15

apanteleev:
But we do not require any features from the latest version. The version specified in the readme and in CMakeLists.txt is the _minimum_ version you should need to build; any later version is OK.

PR #10, opened 2019-06-10
vkpt: fix type definition from auto(which is storage specifier in C) to corresponding types

a1batross:
Looks like this code was copy-pasted from C++11 code. Luckily, C99 allows to not specify the type but it would default to `int`.

### DIFF
--- a/src/refresh/vkpt/device_memory_allocator.c
+++ b/src/refresh/vkpt/device_memory_allocator.c
@@ -144,7 +144,7 @@ int create_sub_allocator(DeviceMemoryAllocator* allocator, uint32_t memory_type)
 	}
 #endif
 
-	const auto result = vkAllocateMemory(allocator->device, &memory_allocate_info, NULL, &sub_allocator->memory);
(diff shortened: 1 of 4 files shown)


apanteleev:
This makes sense, thank you. But it's better to declare the type of all memory flags explicitly as `VkMemoryPropertyFlags` - please update and I'll merge.

a1batross:
Yeah, indeed. Updated. :)

a1batross:
@apanteleev 
off-topic:
I don't understand, vkpt uses tabs or spaces? Yeah, it's topic for holy war but it's would be better to clarify. 

Because for me code looks like [this](https://pp.userapi.com/c849220/v849220903/1b8e1d/mZC0jQg5_zY.jpg). As it's kinda global change, I don't want to out of sync from this repo to effectively merge patches in future.

apanteleev:
Merged, thank you!

Regarding tabs vs. spaces: it's inconsistent, sorry about that. The code has been written by different people using different settings, and nobody cared enough to set a standard or fix it. I usually try to keep the formatting and code style of the file I'm editing though. And it looks reasonable with 4 spaces/tab.

PR #35, opened 2019-07-13
Action mod scope scaling for ch_scale

stan0x:
fixxes scope scaling action quake 2.


apanteleev:
Please provide instructions how to test this change, i.e. how to run the action quake 2 mod with Q2RTX and use this.

stan0x:
This option makes it possible to scale down the scope for SSG sniper rifle in action quake 2 mod.

PR #41, opened 2019-07-25
Make setup more robust to script order.

CKWarner:
There's no reason why this script has to fail if it's run before q2rtx.sh. mkdir -p won't give an error if that directory already exists.


liam-middlebrook:
@apanteleev this looks good to me, can you merge it?

PR #65, opened 2020-02-18
Add SPDX License Identifier

EwoutH:
Added SPDX License Identifier compliant to https://spdx.org/ids-how

### DIFF
--- a/build_materials.py
+++ b/build_materials.py
@@ -1,15 +1,16 @@
 # Copyright (C) 2019, NVIDIA CORPORATION. All rights reserved.
-# 
+# SPDX-License-Identifier: GPL-2.0-or-later
+#
 # This program is free software; you can redistribute it and/or modify
 # it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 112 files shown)


apanteleev:
Thank you for the contribution. I don't mind merging it, but could you please clean up the commit to avoid changing random whitespace?
Also, why do you think that these identifiers are necessary?

EwoutH:
A SPDX identifier makes the license precise, concise and travels with files, which can be an advantage since this is the RTX version is licensed differently from the original version and it's game data.

My text editor removes trailing white spaces by default. Shall I open a separate PR for the trailing whitespaces?

apanteleev:
Concise - yes, but the existing license plaques in every file cover all other aspects that you mentioned. And the license type is stated right there in the first paragraph of the plaque, after the new SPDX header. Besides, this project is not licensed differently from the original Quake 2, Q2PRO, or Q2VKPT: all use GPLv2.

apanteleev:
Old and unnecessary PR, closing.

PR #69, opened 2020-03-29
Make initialization failures fatal

Joshua-Ashton:
This prevents the user launching the game without the prerequisite assets [ie. bluenoise, etc].

This bit me so I figured I'd make this more verbose and explicit.


Joshua-Ashton:
Ah bugger I PR'ed master

PR #72, opened 2020-03-29
Fix some Vulkan validation errors

Joshua-Ashton:
Sometimes we have no particles so we shouldn't perform a degenerate (size = 0) copy as this is forbidden by the spec.


apanteleev:
Merged through rebase. Thank you for the contribution!

PR #73, opened 2020-03-29
Migrate to VK_KHR_ray_tracing

Joshua-Ashton:
This patchset will move Q2RTX to use VK_KHR_ray_tracing.

I'm making this as a draft PR because it has a dependency on #72 being merged in and the extension is currently provisional/beta [and probably also needs a good bit of reviewing.]

When/if #72 gets merged in I will rebase this upon that.

The stuff unique in this PR is from `Update Vulkan headers` henceforth.

Depends on #72.


Joshua-Ashton:
If you want a nice way to review just the commits here not including the header updates etc:

https://github.com/NVIDIA/Q2RTX/compare/b9decbe303edfe3db89d45d254f5ff410e299a37...Joshua-Ashton:master

Joshua-Ashton:
The prerequisites got merged so I unmarked this as a draft.

apanteleev:
I've updated the vulkan headers in master, now a [submodule](https://github.com/KhronosGroup/Vulkan-Headers) is used for them, so you can safely remove those changes from your branch. I tried, and commits 8f0c309 and 6619df7 can be just skipped. The SPIR-V headers were not even used.

However, I can't merge this yet, at least not into master:

1) The mainline drivers currently do not support Vulkan 1.2, one has to use a beta driver instead. On mainline drivers, the executable fails to even load because `vulkan-1.dll` that ships with the driver does not define some necessary symbols, at least `vkGetBufferDeviceAddress`. This can be worked around by shipping a newer `vulkan-1.dll` built from source with q2rtx. Unfortunately, [Vulkan-Loader](https://github.com/KhronosGroup/Vulkan-Loader) does not provide a shared library build mode for Windows, that would be a better option.

2) On a beta driver (442.88), I am getting a Device Lost error right after loading a map. The menu and cinematics work fine. The debug layer cannot be enabled because the driver reports that it is unsupported.

Joshua-Ashton:
> On a beta driver (442.88), I am getting a Device Lost error right after loading a map. The menu and cinematics work fine. The debug layer cannot be enabled because the driver reports that it is unsupported.

Hm, I was testing this on 442.81.0 and it works fine there. I'll take a look when I get the chance.

OmegaMinus1:
Hello, I was cherry-picking this with master and the 451.98 with the Vulkan SDK 1.2.148.1.
The AnyHit shaders are not engaging so the Explosions, Beams, Particles, Sprites are not processing their hit shaders, also they don't process in the primary hit shaders either. I was thinking the setup was keeping the triangles out of the primary but also not processing at all, then I cloned this pr and built it as is and was the same result so I was thinking the behavior is out side the pr its either the new driver or SDK understand it could be needing updated in the pr but I don't see anything wrong with STB setup. Other than this the game played great and this pr has a very nice fps increase, let me sneak in a S.O.S. in my Cheers!

apanteleev:
Hi @OmegaMinus1 - I actually fixed those issues a while ago but forgot to publish the branch... Here you go: https://github.com/NVIDIA/Q2RTX/tree/khr-ray-tracing
This branch used to work at that time (Jul. 12) on the latest Vulkan beta driver.

OmegaMinus1:
Awesome thanks!

BattleAxeVR:
Just tried the KHR branch and had some issues compiling shaders using the .bat file: the default cmake script sets the vulkan version to 1.1 which doesn't have these new beta extensions. 

Changing it to 1.2 manually in the custom build script in visual studio project settings for the affected shaders works, but I wonder if there's a simpler way to change the glslanvalidator command to use vulkan1.2 everywhere by default. Any tips?

oscarbg:
hope this patch gets updated to final raytracing specs and merged soon to see if it works on amd 6800 gpu with today amd drivers supporting vulkan raytracing finally!

CKWarner:
> hope this patch gets updated to final raytracing specs and merged soon to see if it works on amd 6800 gpu with today amd drivers supporting vulkan raytracing finally!

I hope it gets updated to use the finalised extension so that I can play on my Nvidia card. The vendor-neutral extension _is_ the forward-looking one, and Q2RTX is a flagship Vulkan ray tracing title. Nvidia getting support for the extension into their mainline driver, and getting support for the extension into their own game, would demonstrate that they're taking Vulkan ray tracing seriously.

Joshua-Ashton:
Superceded by master.

PR #78, opened 2020-06-18
fix: add flags for gcc 10 so missing extern definitions still compile.

hrehfeld:
See https://gcc.gnu.org/gcc-10/porting_to.html

> Default to -fno-common
> A common mistake in C is omitting extern when declaring a global variable in a header file. If the header is included by several files it results in multiple definitions of the same variable. In previous GCC versions this error is ignored. GCC 10 defaults to -fno-common, which means a linker error will now be reported. To fix this, use extern in header files when declaring global variables, and ensure each global is defined in exactly one C file. If tentative definitions of particular variables need to be placed in a common block, __attribute__((__common__)) can be used to force that behavior even in code compiled without -fcommon. As a workaround, legacy C code where all tentative definitions should be placed into a common block can be compiled with -fcommon.


apanteleev:
GCC 10 build is fixed in release 1.4.0. Closing.

PR #84, opened 2020-09-20
Khronos validation layer

simonbuchan:
Wanted to get validation going, but from what I can tell, the LUNARG_standard_validation is no longer included in the 1.2 SDK. The new hotness KHRONOS_validation is an easy switch, and has a much more descriptive and helpful message by default. The fancy thing here would be to try in order, but that seems to be a bit too fancy for something that's commented out by default. (Perhaps `#define VKPT_ENABLE_VALIDATION KHRONOS` to get the new layer, if that's wanted?)

This makes booting into a map clean, but there's missing deletes for samplers on quit still, and it's possible there are other paths that are not exercised (cinematics? glass/water etc?)

Suggestions welcome on what to test and any style / etc. things I missed (the `malloc()` usage in particular)


simonbuchan:
Can't reproduce sampler error again (argh!), but did find an issue with the unused texture cleanup (due to the `DESTROY_LATENCY` macro expansion it would only have two frames).

simonbuchan:
Ah the god rays shadow map sampler was leaked on swapchain recreation (eg. going into / out of windowed).

I realized I haven't shown the new validation messages, so here's an example:

```
Validation Error: [ VUID-vkDestroyDevice-device-00378 ] Object 0: handle = 0x16935ce7890, type = VK_OBJECT_TYPE_DEVICE; Object 1: handle = 0x4176e9000000037d, name = god_rays.shadow_sampler, type = VK_OBJECT_TYPE_SAMPLER; | MessageID = 0x71500fba | OBJ ERROR : For VkDevice 0x16935ce7890[], VkSampler 0x4176e9000000037d[god_rays.shadow_sampler] has not been destroyed. The Vulkan spec states: All child objects created on device must have been destroyed prior to destroying device (https://vulkan.lunarg.com/doc/view/1.2.141.0/windows/1.2-extensions/vkspec.html#VUID-vkDestroyDevice-device-00378)
```

simonbuchan:
The last issue I've seen is warnings when windows reports a size of 0x0, which happens when the window is minimized. This one is a bit annoying as `SDL_GetWindowSize()` is lagging behind `vkGetSurfaceCapabilities()`, and I'm not sure I get all the logic in `create_swapchain()` enough right now.

apanteleev:
The validation layer has been updated, and many errors and warnings fixed with the release 1.4.0. Closing this PR, please open a new one if there are any similar issues in the new version.

simonbuchan:
Sounds good!

simonbuchan:
Only thing obvious in a quick look over of the diff on my phone is the texture.c DESTROY_LATENCY is still not parenthesised and thus expands to an incorrect index: it seems this was fixed some other way, but perhaps it's something to check?

apanteleev:
> DESTROY_LATENCY is still not parenthesised 

Wow, how did I miss that. Will fix soon, thank you!

PR #90, opened 2020-12-02
Use a file reader callback with tinyobj parser

gunayorbay:
tinyobj no longer uses the previous input arguments including an in memory file buffer.


apanteleev:
Rejected because Q2RTX references a specific commit in the tinyobjloader repository which has the necessary API, and the game builds and runs as-is. Plus, the suggested change replaces `FS_LoadFile` with `fopen` which doesn't support reading package files.

PR #109, opened 2021-03-19
Optionally build with IPO (LTCG)

res2k:
Comparing timedemos of an LTCG build vs the Steam release I get a few percent more FPS if I enable LTCG (and no other changes) - free performance! (A little.)

Still, be conservative and leave it default "off" for now.


apanteleev:
I didn't see any perf improvement from LTCG with this change, but maybe there will be some on systems with slower CPUs. Doesn't hurt to have it. Merged.

PR #110, opened 2021-03-19
Reintroduce 'full screen blend'

res2k:
One of the effects that's currently "lost" in RTX mode is the "full screen blend", where a color is mixed into the final image (literally, an alpha blended screen-covering rectangle). It's used for certain in-game events, eg damage received, item pickups, active effects (quad damage) - that has certainly value for the player. It's also used to give everything a murky color when underwater (so, for atmosphere of a sort).

I added the "full-screen blend" into the tone mapping step, after the final image, but before the tonemapping.
The motivation to add it there, and not the "final_blit" shader, was that I guess making it work with colors not mapped to [0,1] would be more useful if proper HDR output is ever added.


apanteleev:
Merged with a couple of improvements/tweaks. Thank you!

Issue #77, opened 2020-05-13
Allow Lighting from ent files


aepod:
If we could have lights here, we would be able to fix maps without having to edit/recompile. Many old maps are difficult to play on because of darkness. 

I am looking at the Lokis Minions maps mostly, as it has a pretty active community still.

https://photos.app.goo.gl/Jbipn1TMjNTu2cfo9?fbclid=IwAR3d8OmFVROvIBdOit6AYPt18aocNDhb2W2DUxy23WrEdnX0Fs1XM4rSdTs

Here in the stills you can see dark areas. I also used an old demo from 1999, and the results look stunning, but you will notice the dark areas near the lava would make it near impossible to "play" this way.
https://www.youtube.com/watch?v=prNwu09khnA

```
{
"origin" "1448 968 -96"
"light" "100"
"classname" "light"
}
```
This is an example of a light from the ent file, new classname could be added for other types of lights, to accommodate quickly fixing older maps.  I suggest doing different types of lights with some ability to change parameters.

juks01:
The other way could be to update the client to support those original light sources (entity lights?).
It would probably fix the problem on almost every mod and map.

aepod:
The original q2kpt had this issue and there was even a potential path forward outlined here:
https://github.com/cschied/q2vkpt/issues/21#issuecomment-456750791

savvykms:
I actually forked q2rtx specifically to take a swing at this:
https://github.com/savvykms/Q2RTX/tree/entity-lights

My current build doesn't seem to be functioning properly, however. I was also hasty in my design, so it's likely not the best approach. I'm just roughly mashing together the normal quake2 game logic's way of parsing the entity string from the loaded bsp_t object, caching it, and rendering it each frame. These entity lights would affect the static lighting in the map during compilation, and qrad3 (or similar) could have its settings adjusted when run, meaning each map could have different values - this is especially noticeable for maps created 2005+ where people really started tinkering due to better hardware.

I haven't addressed conical lights (e.g. when one targets another entity), "dynamic" entity lights (e.g. light styles with flickering etc), lightramp entity logic, lack of IR functionality, or significant changes to some dynamic lights that certain mods (e.g. gloom) used for lighting around player entities in certain conditions,

savvykms:
[...]

PR #111, opened 2021-03-25
Add support for BSP "LIGHT" surfaces

res2k:
For lighting, Q2RTX predominantly uses light-emitting surfaces. The amount of light a surface should emit is determined by an entry in the materials list and an "emissive" texture associated with a material. Material definitions and, if appropriate, emissive textures are provided for basically all baseq2 materials.
Unfortunately, if an emissive texture is not present, a surface won't emit light. This is a problem for maps that use materials beyond the baseq2 set - ie from mods or mission packs (see also #77).
Fortunately, the original light information is still present: the `SURF_LIGHT` flag indicates that a surface should emit light.
With this information "old" maps could be lit much more like they were intended to.

This patch adds a check for `LIGHT` surfaces and the ability to synthesize an appropriate light-emitting material.
An emissive texture is guessed from the diffuse texture by taking the "brightest" pixels. This looks okay for textures in the typical Quake 2 "wall light" style. (An earlier approach was to use the diffuse texture only. While useable, it's not the most attractive look.)

Additionally, while trying out the changes, I noticed:
* Quite rarely there's a surface with a `LIGHT` flag, and it's texture has a `materials.csv` entry, but it lacks the emissive texture (happens in eg `base3`). This probably indicates an "artwork problem" - missing emissive texture - that results in surfaces not emitting light, though they were originally intended to.
* Water and slime are supposed to faintly emit light! This makes a visible difference in some cases (eg `base1` cellar). To see the difference set `pt_enable_surface_lights_warp` to 1. I left it off by default since mainly because it's a hack, having water emit light without light textures. That "proper" way is also available (set both `pt_enable_surface_lights` and `pt_enable_surface_lights_warp` to 2), but the result looks quite weird/jarring/unappealing.<br/>FWIW, the faint "water light" might very well be relevant in certain dark areas, so it might be good to find a way to have it back (without hacks).

### DIFF
--- a/inc/refresh/images.h
+++ b/inc/refresh/images.h
@@ -107,6 +107,7 @@ extern uint32_t d_8to24table[256];
 // these are implemented in src/refresh/images.c
 void IMG_ReloadAll();
 image_t *IMG_Find(const char *name, imagetype_t type, imageflags_t flags);
+image_t *IMG_Clone(image_t *image);
 void IMG_FreeUnused(void);
 void IMG_FreeAll(void);
 void IMG_Init(void);
--- a/inc/refresh/images.h
+++ b/inc/refresh/images.h
@@ -107,6 +107,7 @@ extern uint32_t d_8to24table[256];
 // these are implemented in src/refresh/images.c
 void IMG_ReloadAll();
 image_t *IMG_Find(const char *name, imagetype_t type, imageflags_t flags);
+image_t *IMG_FindExisting(const char *name, imagetype_t type);
 image_t *IMG_Clone(image_t *image, const char* new_name);
 void IMG_FreeUnused(void);
 void IMG_FreeAll(void);
--- a/inc/refresh/images.h
+++ b/inc/refresh/images.h
@@ -107,7 +107,7 @@ extern uint32_t d_8to24table[256];
 // these are implemented in src/refresh/images.c
 void IMG_ReloadAll();
 image_t *IMG_Find(const char *name, imagetype_t type, imageflags_t flags);
-image_t *IMG_Clone(image_t *image);
+image_t *IMG_Clone(image_t *image, const char* new_name);
 void IMG_FreeUnused(void);
 void IMG_FreeAll(void);
 void IMG_Init(void);
--- a/inc/refresh/refresh.h
+++ b/inc/refresh/refresh.h
@@ -202,6 +202,7 @@ typedef enum {
     IF_NEAREST      = (1 << 7),
     IF_OPAQUE       = (1 << 8),
     IF_SRGB         = (1 << 9),
+    IF_FAKE_EMISSIVE= (1 << 10),
 
     // Image source indicator/requirement flags
     IF_SRC_BASE     = (0x1 << 16),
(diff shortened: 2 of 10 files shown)


res2k:
Hey @apanteleev, what's your opinion on this?
For one, I'm unsure whether PRs like this, to "make old stuff work", is something you consider worthwhile for Q2RTX, or if you want to focus on "baseq2"-related stuff only...

aepod:
This is an amazing PR, and will fix many Maps throughout many mods. For instance Lokis Minions almost all maps use emissive textures as lights, so in RTX they are far too dark. Without going through and manually updating maps this is impossible to fix. This fix solves that. There are literally thousands of maps that will possibly fixed by this PR.

Kudos!

apanteleev:
> Hey @apanteleev, what's your opinion on this?

Well, in general, I wouldn't mind merging this. But one thing I dislike about the implementation is that there seems to be a significant amount of image processing being done on the CPU at map load time. Maybe it's insignificant, but I feel like our loading process is slow enough already, with all the image extraction from .pkz and decoding being done in a single thread. Have you measured the impact of this processing on load times?
(Actually, it probably doesn't matter for baseq2 because the synthesis is disabled by default through the new cvars, right?)

res2k:
Well, while the main cvar is "on" by default, it shouldn't matter too much for baseq2, as the mechanism doesn't do any image processing if there's a material.csv entry for the texture, which (to my knowledge) all baseq2 wall textures have.

Yes, image processing on the GPU is quite sensible. But the CPU approach more easily fits into the current "upload a bunch of pixels" model, while a GPU processing component will need some enhancements to the texture uploading mechanism.
But in general, it's certainly an enhancement I'm interested in doing.

Thanks for your feedback. It's good to know directions to get my ideas shared with the world :)

(Unfortunately I don't have hard numbers on the load time difference. I can't quickly produce some right now either - there's no "map load time" measurement built in, and subjectively, the difference is too small to measure manually. Perhaps reporting a "map load time" is a worthwhile addition in itself.)

apanteleev:
I've tested this PR with the current `master`, and it works... in some cases. Not very reliably. Like, in Ground Zero it mostly doesn't work, I think because the 215 unit threshold for brightness is too high for many of the textures. Maybe introduce a cvar? Also, ignoring the baseq2 materials is not always a good solution.
I'll take another look at the problem later.

res2k:
My go-to test was mostly some maps of xatrix (The Reckoning), and it seemed fined there... but yeah, it's very texture-dependent (and also subjective).

Yes, a cvar would a good way to quickly see what effects different thresholds have.

What would probably be _best_ would be to have a database that allows tweaking of the threshold (maybe other parameters) per texture.
But with the sheer amount of mods, maps out there... well, I suppose a robust automatic solution would be much more manageable ;)

In an earlier iteration I had this:
`float threshold = max_lum * 0.68f;`
...perhaps an approach of "guessing" the threshold could produce appealing results for a wider range of textures.

BTW, did you mean this: https://github.com/res2k/Q2RTX/commit/2415da21620e07aeada906780097512ac7689efd with "ignoring baseq2 materials"?
Generally, if the change in question is from a PR of mine, I can probably explain the reasoning behind it ;)

res2k:
@apanteleev Also check this commit: 937fb7757f622155b88c1edafa1252b29750cf9d, may affect your experience with Ground Zero.

Also, if which levels in Ground Zero did you look at? I'd like to check them out as well.

apanteleev:
> which levels in Ground Zero did you look at?

Just the first one, `rmine1`. It was pretty dark.

res2k:
> Just the first one, `rmine1`. It was pretty dark.

Please try again with the latest change, this improves lighting there quite dramatically.

PR #112, opened 2021-03-31
Add autocompletion support for ray_tracing_api cvar

res2k:
A small convenience when switching around for testing.


apanteleev:
Merged, thank you!

PR #116, opened 2021-04-13
Improve support for old mods

res2k:
...such as those you'd find in the GOG version of Quake II.

* A 32-bit dedicated server build is provided, so old `gamex86.dll` game libraries can be used.
* Inhibition of other config files when `autoexec.cfg` is present was removed. That behavior broke mods that rely on `default.cfg` being executed.
* Texture and model asset loading was adjusted so a baseq2 "better quality" asset doesn't automatically override assets from the game. This is allows mods to actually replace models, textures and images.

Limitations:
* Since it's a dedicated server, only multiplayer games work.
* No Linux support. It should probably work, but I don't really have any i386 mods to test with.

But hey, it's enough to play against some bots -
start a server with `q2rtxded_x86.exe +set game eraser`, set `bot_num` to some value in the terminal, and connect with `q2rtx.exe +connect localhost`.

### DIFF
--- a/inc/common/files.h
+++ b/inc/common/files.h
@@ -87,6 +87,8 @@ typedef struct file_info_s {
 #define FS_Mallocz(size)        Z_TagMallocz(size, TAG_FILESYSTEM)
 #define FS_CopyString(string)   Z_TagCopyString(string, TAG_FILESYSTEM)
 #define FS_LoadFile(path, buf)  FS_LoadFileEx(path, buf, 0, TAG_FILESYSTEM)
+#define FS_LoadFileFlags(path, buf, flags)  \
+                                FS_LoadFileEx(path, buf, (flags), TAG_FILESYSTEM)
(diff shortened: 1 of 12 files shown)


apanteleev:
Merged with some changes. Specifically, I didn't like the idea of calling the CMake generator from a CMake file, so I changed the configuration to work for both 64-bit targets (full build) and 32-bit targets (server and game only build).

Thank you for the contribution!

res2k:
The motivation behind the nested CMake call was that 64-bit client, 64-server, but also 32-bit server could be built with a single action; there was no need to manually set up a separate build directory or copy the 32-bit server to the right place...

apanteleev:
I understand the idea, but I believe that CMake should be used for its primary purpose only, i.e. generating solutions or makefiles. When the same system is used to do other things, such as pulling dependencies or building things, that complicates the configuration process, makes it more fragile and less flexible.

PR #122, opened 2021-06-09
Remove MAX_SWAPCHAIN_IMAGES limit for XWayland

Joshua-Ashton:
A well-behaved Vulkan application should not care about the number of swapchain images exposed by vkGetSwapchainImagesKHR.

In XWayland (right now), there is no IMMEDIATE mode, only MAILBOX, which is exposed as IMMEDIATE.

However, this means that the max number of swapchain images exceeds what would normally be expected for typical IMMEDIATE implementations under X11.

With this change I can run Quake 2 RTX under XWayland on my Radeon VII.
This will also likely be needed by NVIDIA following the release of their 470 driver which enables [X]Wayland support.

Signed-off-by: Joshua Ashton <joshua@froggi.es>

### DIFF
--- a/src/refresh/vkpt/draw.c
+++ b/src/refresh/vkpt/draw.c
@@ -52,7 +52,7 @@ static VkPipelineLayout        pipeline_layout_final_blit;
 static VkRenderPass            render_pass_stretch_pic;
 static VkPipeline              pipeline_stretch_pic;
 static VkPipeline              pipeline_final_blit;
-static VkFramebuffer           framebuffer_stretch_pic[MAX_SWAPCHAIN_IMAGES];
+static VkFramebuffer*          framebuffer_stretch_pic = NULL;
 static BufferResource_t        buf_stretch_pic_queue[MAX_FRAMES_IN_FLIGHT];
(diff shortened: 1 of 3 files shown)


xcom169:
Hello All! This game is only XWayland right now, or can be run with pure Wayland?

Joshua-Ashton:
It works on native wl

xcom169:
Anything extra is needed for the Wayland mode? For me it looks XWayland is used by default on my Manjaro/Gnome rig.

Joshua-Ashton:
SDL_VIDEODRIVER=wayland

xcom169:
Thanks, but somehow it just generates an error message:
[image: image.png]

xcom169:
Sorry, the image is here: 
image

res2k:
Okay, the issue is a combination of (a) Q2RTX on Linux shipping it's own copy of SDL and (b) that copy not being built with Wayland support. (See eg here: https://github.com/NVIDIA/Q2RTX/runs/5505535453?check_suite_focus=true, the build-linux-x64/Configure output.)
I think this is a genuine issue and should probably get it's own GH issue instead of hijacking this (closed!) PR for further discussion.

PR #146, opened 2021-09-20
Change 'procedural skipping' flag into compile time constant

res2k:
This picks up on a comment in `path_tracer_rgen.h`, where it says `skip_procedural` "should be a compile-time constant".
Now it is.

### DIFF
--- a/src/refresh/vkpt/path_tracer.c
+++ b/src/refresh/vkpt/path_tracer.c
@@ -1206,28 +1206,21 @@ vkpt_pt_destroy()
 VkResult
 vkpt_pt_create_pipelines()
 {
-	VkSpecializationMapEntry specEntry = {
-		.constantID = 0,
-		.offset = 0,
-		.size = sizeof(uint32_t),
-	};
-
-	uint32_t numbers[2] = { 0, 1 };
-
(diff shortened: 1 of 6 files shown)


apanteleev:
Thanks, but... do you see any actual speedup from these changes? What the comment really meant is, the value of that `skip_procedural` parameter should be a literal constant, `true` or `false`, and not something from UBO for example. This way, the compiler should be able to figure it out statically and apply constant folding, to the same effect as these specializations.

In any case, there are some pending changes to transparency handling that make this pull request conflicting. They're not quite ready yet, but you can take a look here: https://github.com/apanteleev/Q2RTX/tree/fog2

res2k:
Speed up ... yes and no. In "query" mode, timedemoing demo2 runs 0.2-0.3fps faster in 4k, in "pipeline" mode its 0.1 fps slower.
Either way, nothing to write home about, really.

FWIW, `skip_procedural` is already a compile-time constant in the sense you mentioned - all calls to `trace_ray` have a literal `true` or `false` argument for `skip_procedural`.
Would be interesting to look at the generated SPIRV to see if the compiler does anything with that.
Maybe I'll do that someday 🙃

Issue #83, opened 2020-09-18
HDR output


simonbuchan:
Having just got an HDR-capable monitor, I of course went down the list of games to see what supported it - to my mild surprise, Q2RTX wasn't one of them!

After figuring out the very basics of doing HDR in Vulkan on my own toy project, I took a swing at the source to see how plausible it would be to add it to this: to my surprise, simply adding `VK_FORMAT_R16G16B16A16_SFLOAT` to `acceptable_formats` in `create_swapchain()` and setting `\tm_enable 0` to get the raw HDR data on screen was enough to get some pretty damn good results! Apparently the physically accurate lighting work already done mean that without the tonemapping crushing the range you get pretty good results when in the right environments.

Strictly though, `VK_EXT_SWAPCHAIN_COLOR_SPACE_EXTENSION_NAME` should also be added to `vk_requested_instance_extensions`, which for my screen means it uses the declared surface format color space of `VK_COLOR_SPACE_EXTENDED_SRGB_LINEAR_EXT`. Not exactly sure what the interpretation is without it, certainly with the extension enabled it is darker, implying that it is no longer applying sRGB gamma correction, as implied by the color space name (presuably the SRGB is the value of the primaries). I'll need to look into what the right thing here is, since `IMG_TAA_OUTPUT` should be linear already... I might just be confused.

So far the main issues I have hit are:
* Obviously, without *any* tonemapping the results are not optimal, morning and day times the sky is blown completely to white on my fairly typical HDR monitor, which is probably accurate, but not ideal. Some readable areas with `\tm_enable 1` become uncomfortably dark, though still visible. I'm not sure where the state of the art is here, but from what I've seen one approach is roughly the same as the existing logic, but with a configurable output range, rather than 0-1 to avoid monitor clipping, and a configurable paper white brightness to scale the exposure to match (Though everyone seems to be doing it differently right now). Maybe I'll take a stab at this, the tonemapping shader doesn't seem too confusing, but my early blind stabs were pretty unsuccessful at getting anything good out (I can get HDR values just by disabling the clamp at the end of the apply and using "out of range" values for `tm_*` cvars, but they all looked terrible - I probably just need to sit down and re-figure out the maths)
* With resolution scaling, there's a solid black line (one input pixel thick) around the blown-to-white sky and other lights, due to the lanczos filtering overshooting with the huge value step. Not sure how much of an issue this will be with "proper" tone mapping to avoid monitor clipping.
[...]

PR #159, opened 2021-11-04
HDR support

res2k:
This builds upon the basic approach outlined by #83 ("pick float16 surface format") and updates various systems for HDR.

Tonemapping: This tweaks the current algorithm to disable the "knee" for bringing things into range and clamping/sRGB conversion. Extending the brightness range is done by applying a (user-configurable) scale after tonemapping; seems to work and look well enough.

Screenshots: They can be written in the Radiance `.hdr` format. The main motivation for using this was simplicity: STB comes with an HDR writer. Viewers are available online (eg https://viewer.openhdr.org/) and standalone (eg https://github.com/wkjarosz/hdrview).

Other tweaks include:
* Adjustments to FSR shaders to handle HDR values
* User-configurable UI brightness

### DIFF
--- a/baseq2/q2rtx.menu
+++ b/baseq2/q2rtx.menu
@@ -96,6 +96,7 @@ begin hdr
     blank
     range -f "%.0f nits" "peak brightness" tm_hdr_peak_nits 100 2000 10
     range -f "%.0f nits" "UI brightness" ui_hdr_nits 100 2000 10
+    range -f "%.0f%%" "Saturation adjustment" tm_hdr_saturation_scale 0 200 5
 end
 
 begin photomode
--- a/baseq2/q2rtx.menu
+++ b/baseq2/q2rtx.menu
@@ -88,6 +88,16 @@ begin resolution
     range -f "%.0f%%" "fixed resolution scale" viewsize 25 200 5
 end
 
+begin hdr
+    background #212F3CD0
+    style --compact
+    title "HDR Settings"
+    toggle "Enable HDR" vid_hdr
+    blank
+    range -f "%.0f nits" "peak brightness" tm_hdr_peak_nits 100 2000 10
+    range -f "%.0f nits" "UI brightness" ui_hdr_nits 100 2000 10
+end
+
 begin photomode
     background #212F3CD0
     style --compact
@@ -120,6 +130,7 @@ begin video
 
     ifeq vid_rtx 1
         action "resolution scaling options..." pushmenu resolution
+        action "HDR options..." pushmenu hdr
         blank
         range -f "%+.1f EV" "exposure bias (brightness)" tm_exposure_bias -5 0 0.1
         range -p -f "%.0f%%" "contrast" tm_reinhard 0 1 0.1
--- a/cmake/compileShaders.cmake
+++ b/cmake/compileShaders.cmake
@@ -2,6 +2,8 @@ set(SHADER_SOURCE_DEPENDENCIES
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/asvgf.glsl
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/brdf.glsl
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/constants.h
+    ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/fsr_easu.glsl
+    ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/fsr_rcas.glsl
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/global_textures.h
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/global_ubo.h
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/god_rays_shared.h
--- a/cmake/compileShaders.cmake
+++ b/cmake/compileShaders.cmake
@@ -4,6 +4,7 @@ set(SHADER_SOURCE_DEPENDENCIES
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/constants.h
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/fsr_easu.glsl
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/fsr_rcas.glsl
+    ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/fsr_utils.glsl
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/global_textures.h
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/global_ubo.h
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/god_rays_shared.h
--- a/inc/refresh/images.h
+++ b/inc/refresh/images.h
@@ -127,6 +127,7 @@ void IMG_MipMap(byte *out, byte *in, int width, int height);
 extern void (*IMG_Unload)(image_t *image);
 extern void (*IMG_Load)(image_t *image, byte *pic);
 extern byte* (*IMG_ReadPixels)(int *width, int *height, int *rowbytes);
+extern float* (*IMG_ReadPixelsHDR)(int *width, int *height);
 
 #endif // IMAGES_H
 
--- a/inc/refresh/refresh.h
+++ b/inc/refresh/refresh.h
@@ -291,6 +291,7 @@ extern void    (*R_ModeChanged)(int width, int height, int flags, int rowbytes, 
 extern void    (*R_AddDecal)(decal_t *d);
 
 extern qboolean (*R_InterceptKey)(unsigned key, qboolean down);
+extern qboolean (*R_IsHDR)();
 
 #if REF_GL
 void R_RegisterFunctionsGL();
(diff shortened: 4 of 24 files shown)


Rroff2:
Had a quick play with it - on my Windows 10 machine RTX3070 with Philips 436M6 it works without any problems and looks amazing - had to fiddle with the tone mapper a little, mostly on custom maps.

On my Windows 7 system without a HDR display it correctly errors out with "HDR was requested but no supported surface format was found." without any problems.

Not much more to add than that at this point - HDR is a good bit beyond my working knowledge from a technical perspective.

res2k:
I'm curious, what tone mapper fiddling did you do/what were the issues on custom maps?

Rroff2:
It is more a sun problem than tone mapper really - on some custom maps they have completely opposite behaviour with the sun overpowering the scene depending on whether the tone mapper is enabled or disabled - not sure what is going on there as it is probably a map design issue rather than renderer problem as such.

I'm really liking the look with HDR on and TM disabled in the stock maps though - it brings back a lot of the feel of Quake 2 as per how I remember it in my mind with the bright lights and deep shadows.

res2k:
Thanks for testing.
It could be possible the custom maps "always" had the issue of an overpowering sun, but it was less noticeable due to the mapping into a more limited (SDR) brightness range...
If you don't know already, Q2RTX supports per-map config file, so if some `tm_XXX` cvar mitigates the issue, this could be a way.

Generally, while just "disable tonemapping" looks surprisingly good, I found to be not without issues either. Light textures on walls (high brightness values) are quickly blown out to white, while dark areas would end up to be too dark to see much. 
To some extent this is a question of personal taste, but IMO the latter is also a playability issue. (Though, it's possible the issues at the dark or bright end could be display-specific.)

Ah well. At the end it's a bit of a subjective matter and a bit of a balancing act.
But there can always be more tweaks later 🙃

Rroff2:
You are probably right - level design for Quake 2 was always a compromise between physical accuracy and what *looks* right within the limits of the technology.

On my Philips 436M6 with TM disabled it works really well on the stock maps so probably is somewhat display specific - the display seems to have have slightly elevated blacks and the lights aren't blown out enough to ruin it resulting in a great mix of deep shadows and vibrant, if a little stark, lighting.

Great work.

Rroff2:
Maybe an error on my end - build environment or because I manually merged the code and some of the changes in images.c are a little difficult to follow - but I'm getting an error ("Couldn't write screenshots/quake165.jpg: Library error") on trying to take a screenshot without HDR enabled coming from "Com_EPrintf("Couldn't write %s: %s\n", buffer, Q_ErrorString(ret));" in make_screenshot in images.c. But it still writes out the screenshot fine.

Tried cleaning out the screenshots folder, changing resolution and FSR settings, etc. just in case, clean install of the game as per official 1.5 version works fine.

apanteleev:
I gave this branch a try... Before I describe my findings, I'd like to note that I have no experience working with HDR displays whatsoever, and I don't have a proper HDR monitor. I was testing the game on a 5-year-old Samsung 4K HDR TV that was in "game mode".

In HDR mode, the picture looks more washed out than in SDR. I don't know if the TV is doing any improvements to it in SDR mode; I suppose it does, because I get more contrast and vibrance this way. Tried playing with the tone mapper settings, and setting `tm_reinhard 1` with the max nits at something like 1200 seems to be the best option.

I also found that the max nits setting affects the rendering in SDR mode, too - it probably shouldn't.

Screen pictures taken with a phone, for comparison.

HDR mode:
hdr-mode

SDR mode:
sdr-mode

Rroff2:
If you are using a TV make sure full range RGB is enabled.

Rroff2:
HDR w/ TM enabled looks like this on my screen - phone camera doesn't do it justice:


Can do some interesting things with HDR + fog and fiddling with the exposure bias:

res2k:
> In HDR mode, the picture looks more washed out than in SDR. I don't know if the TV is doing any improvements to it in SDR mode; I suppose it does, because I get more contrast and vibrance this way. Tried playing with the tone mapper settings, and setting `tm_reinhard 1` with the max nits at something like 1200 seems to be the best option.

Probably all displays do some processing in SDR mode ... I can also notice some differences on my HDR display between modes, though not to the extend visible on your photos. For me, HDR tends to have a more "neutral" appearance with slightly less color saturation.

One idea may be to check how other apps/games look...
If they have similar appearance differences across the board it would suggest to me that there isn't much to be done about "TV meddling".
On the other hand, if they do look similar in SDR and HDR, it suggests there's more we can do in code. (Different color space? Provide "HDR metadata"?)

Things perhaps to try:
* Chrome can play HDR YouTube videos
* VLC can play HDR source material
* Other games

> I also found that the max nits setting affects the rendering in SDR mode, too - it probably shouldn't.

Hm, that's odd. `tm_hdr_peak_nits` usage is only used in "if(HDR)" branches.
Can you describe how this affects the rendering?

I also noted that switching `vid_hdr` on the fly doesn't quite work right.
E.g dynamically changing from 1 to 0 results in an incorrectly tonemapped image. (Though the other way round seems ok.)
While I'll certainly need to take another look there, it's possible that some issue with dynamic `vid_hdr` switching may affect how things turn out for you.

draak567:
I used to have a lower-end Samsung "4K HDR" TV from back in 2015 or 2016 and it washed and greyed out HDR content like in that image. Unless you paid thousands of dollars back then, and the TV was advertised as "HDR1000", the best you got with Samsung's lower-end TVs was "fake HDR" where it tone-mapped HDR to SDR, poorly. Even with their highest-end TVs, the firmware didn't support HDR10+ or HLG out of the box (so forget HDR YouTube or some select streaming content providers).

For what it's worth, I tried this patch on an Acer Predator X35 (1000 nits) and it works well with the stock maps. If you turn HDR off in the game without restarting it, the colors are wrong, but other than that, I didn't see any major issues. 

Only nit-pick I might have is that the dark areas are a bit brighter than in SDR by default (though that helps with the visibility in many places and you can fiddle with the maximum nits setting to mitigate it somewhat) and color saturation seems a tad lower than in SDR (not sure what can be done about that). 

Turning tone-mapping off produces a gritty image with deep blacks and dark corners and bright lights, but it does blow out the outside sky.

The image in SDR mode was identical with and without this patch on my screen.

res2k:
Hey @draak567,
> Only nit-pick I might have is that the dark areas are a bit brighter than in SDR by default 
You could also try to fiddle a bit with the `tm_noise_blend` cvar. Lowering it from it's default 0.5 turns dark areas somewhat darker.

> and color saturation seems a tad lower than in SDR (not sure what can be done about that).
I suspect that's due to image processing differences in the hardware. On my display can I also notice some differences in saturation between SDR and HDR.

draak567:
> You could also try to fiddle a bit with the `tm_noise_blend` cvar. Lowering it from it's default 0.5 turns dark areas somewhat darker.

You're right, lowering that does help, though what has the most effect on my monitor is lowering the maximum HDR nits setting by a bit.
One thing is that I've changed the gamma and contrast settings in SDR on this monitor from the default (to increase contrast) and the HDR modes don't have gamma or contrast control at all. If I change the monitor's SDR settings back to default, the brightness of dark areas in SDR and HDR is much more comparable.

> I suspect that's due to image processing differences in the hardware. On my display can I also notice some differences in saturation between SDR and HDR.

I think that's definitely possible, the saturation is also a bit affected in other games I've tried. Most AAA games released in recent years support HDR but results are variable. Some games actually look much better in SDR than HDR and no amount of fiddling with the game's HDR settings help.

It's also possible to just increase the saturation (sometimes called "digital vibrancy" or similar) in the control panel for the video driver. I've had to do that manually for some games.

res2k:
@draak567, @apanteleev, as a bit of an experiment, I added a "Saturation adjustment" option.
Perhaps this allows to adjust the HDR picture saturation to something more pleasing to you (though, TBH, the adjustment formula isn't very sophisticated).

Rroff2:
I feel that the saturation isn't so much a technical problem but that the original maps just weren't designed with this kind of lighting in mind (despite using radiosity, etc.) and the loss of input from some of the sources like point lights which were used to produce something aesthetically pleasing with the limits at the time. With custom mapping it is possible to get much better lighting and saturation results than the vanilla maps.

apanteleev:
I gave this PR another try, after the latest changes like saturation and after disabling as many "image improving" settings in the TV as I could. With saturation adjusted to something like 120-130%, the HDR mode does look better than SDR. The issue with the max nits setting affecting the SDR image has also been fixed. The code looks good, too.

One final question: how difficult is it to implement an HDR on/off switch without vid_restart? Looks like we just need to destroy and create the swap chain again, like what happens on window resize.

res2k:
> One final question: how difficult is it to implement an HDR on/off switch without vid_restart? Looks like we just need to destroy and create the swap chain again, like what happens on window resize.

Actually, sounds like it shouldn't be too difficult. I'll have a look.

res2k:
> One final question: how difficult is it to implement an HDR on/off switch without vid_restart? Looks like we just need to destroy and create the swap chain again, like what happens on window resize.

@apanteleev, implemented now.

apanteleev:
Excellent. I've tested this a bit more, switching works well, and it's easier to compare HDR vs. SDR this way. Sometimes the colors are just different, and not really correctable with the saturation adjustment, but... maybe that can be tweaked later.

Merged.

Thank you for another valuable contribution, @res2k!

Rroff2:
Some of the problem may be Windows 10 - its HDR mode is somewhat rough and ready at the best of times and SDR in the application while Windows 10 has HDR enabled may be doing odd things behind the scenes.

PR #164, opened 2021-11-25
Add an 'rtx' variant of basedir 

res2k:
In the recent LFS poll, there was mentioning of "Game can be built and ran just by cloning the repo, building it and pointing to the original *.pak files." Incidentally, I had laying something around for the "pointing to *.pak files" part...

Currently, the engine knows a "base directory" (cvar `basedir`), with game directories below, which contain files and the game library. This change would add the notion for a "RTX base dir", which can contain additional game directories, itself with data files and (possibly) game libraries.

From the 6074743 message:

> Game library loading was adjusted to check both libdir
> (by default location of Q2RTX executable) and basedir (possibly
> set by user), in that order.
> 
> If basedir_rtx is not set, everything behaves as before.
> 
> basedir_rtx can be set to the special value ".", which means
> "directory of executable".

You can see how this would be used in practice by in ` filesystem.cfg.example`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ save
 */docs
 baseq2/**/*.txt
 !baseq2/*.txt
+filesystem.cfg
--- a/filesystem.cfg.example
+++ b/filesystem.cfg.example
@@ -1,0 +1,4 @@
+// Grab game data from GOG install
+set basedir "C:\Program Files (x86)\GOG Galaxy\Games\Quake II"
+// Look for RTX media next to executable
+set basedir_rtx "."
--- a/inc/shared/platform.h
+++ b/inc/shared/platform.h
@@ -68,6 +68,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #define os_fstat(f, s)      _fstat(f, s)
 #define os_fileno(f)        _fileno(f)
 #define os_access(p, m)     _access(p, m)
+#define os_chdir(p)         _chdir(p)
 #define Q_ISREG(m)          (((m) & _S_IFMT) == _S_IFREG)
 #define Q_ISDIR(m)          (((m) & _S_IFMT) == _S_IFDIR)
 #define Q_STATBUF           struct _stat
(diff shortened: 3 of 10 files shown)


apanteleev:
(Old PR but still)
I don't think we should complicate the file search logic, it can already be confusing. 
Closing; if you or someone else feels that this is a useful change, please repoen and we can discuss.

res2k:
(FWIW, I don't see a "reopen" button, maybe something only you can do?)

apanteleev:
Ok, reopening... I'll give it some more thought.

res2k:
I'll try to add some discussion on the different pieces mentioned in https://github.com/NVIDIA/Q2RTX/pull/164#issuecomment-1846077404.

* Filesystem search directories (2nd base dir)
This is the "heart" of the idea. Luckily, the actual changes to filesystem code aren't very complicated - just some additional ` add_game_dir()` calls.
* `filesystem.cfg` special config file

res2k:
I'll try to add some discussion on the different pieces mentioned in https://github.com/NVIDIA/Q2RTX/pull/164#issuecomment-1846077404.

* Filesystem search directories (2nd base dir)
This is the "heart" of the idea. Luckily, the actual changes to filesystem code aren't very complicated - just some additional ` add_game_dir()` calls.
* `filesystem.cfg` special config file
The usefulness of this depends on whom the "RTX dir" is targeted to. If targeted towards "developers" and "enthusiasts", just specifying the "RTX dir" via command line is probably fine. An additional config file would mostly be a matter of convenience. But if scenarios where adjusting the command line is cumbersome and/or inconvenient should be supported - for example, the Steam install process could be changed from copying files to simply writing a filesystem config -, the extra file makes sense.
* Search for game library in RTX dir
Probably needed - if eg the "base base directory" is pointing to an "original" Quake 2 install, only 32-bit game DLLs would be inside, so there needs to be some logic to pick up the game DLL shipping with Q2RTX.
* Music from RTX dir
Music shipping w/ Q2RTX seems very unlikely, so probably unneeded.

apanteleev:
The project is discontinued; no new features will be implemented here. Closing.

Issue #153, opened 2021-10-04
Correct behavior of "intensity", gl_brightness" and "gl_modulate" in OpenGL


TobiasKarnat:
This backports the patch from Q2PRO by void-995 to Q2RTX.
https://github.com/skullernet/q2pro/pull/144

Most noticeable is that the lava doesn't look washed out any more with intensity != 1.
![1](https://user-images.githubusercontent.com/3251610/135907841-3e446f31-7d9d-4b4f-a7fe-0d51c0f7b80e.png)

![2](https://user-images.githubusercontent.com/3251610/135907860-34cf8578-4ab5-490f-9d63-b2327de6c61a.png)

```
diff -urN a/src/refresh/gl/arbfp.h b/src/refresh/gl/arbfp.h
--- a/src/refresh/gl/arbfp.h	2021-10-04 17:18:06.793235140 +0200
+++ b/src/refresh/gl/arbfp.h	2021-10-04 19:19:37.520254670 +0200
@@ -2,16 +2,48 @@
     "!!ARBfp1.0\n"
     "OPTION ARB_precision_hint_fastest;\n"
 
-    "TEMP ofs, coord, diffuse;\n"
-    "PARAM amp = { 0.0625, 0.0625 };\n"
+    "TEMP offset, coordinate, diffuse;\n"
+    "PARAM amplitude = { 0.0625, 0.0625 };\n"
     "PARAM phase = { 4, 4 };\n"
 
-    "MAD coord, phase, fragment.texcoord[0], program.local[0];\n"
-    "SIN ofs.x, coord.y;\n"
-    "SIN ofs.y, coord.x;\n"
-    "MAD coord, ofs, amp, fragment.texcoord[0];\n"
-    "TEX diffuse, coord, texture[0], 2D;\n"
+    "MAD coordinate, phase, fragment.texcoord[0], program.local[0];\n"
+
+    "SIN offset.x, coordinate.y;\n"
+    "SIN offset.y, coordinate.x;\n"
+
+    "MAD coordinate, offset, amplitude, fragment.texcoord[0];\n"
+    "TEX diffuse, coordinate, texture[0], 2D;\n"
+
     "MUL result.color, diffuse, fragment.color;\n"
     "END\n"
 ;
 
+static const char gl_prog_lightmapped[] =
+    "!!ARBfp1.0\n"
+
+    "TEMP diffuse, lightmap;\n"
+
[...]

PR #166, opened 2021-12-10
Q2pro updates

Paril:
This branch pulls in all of the changes from Q2PRO between c3e4273d800fafbf50883f2a7b97eb7fd0161b74 and 59d0fc43ccecaab878889a2a33393faeab2e7dca (the last commit Skuller had made), bringing Q2RTX up to date with Q2PRO's upstream.

The majority of large changes were tested, but I've not validated that every little detail is correct. Some extra eyes would help.


res2k:
This is quite a large change with 355 commits. This is hard to review ... e.g. I don't think I have the stamina to do this.

However, just from a quick look, it seems the changes range from maintenance without semantic changes (like qboolean -> bool) to e.g. breaking changes (savegame format) - there would probably be different considerations for merging these changes (or not).
My recommendation would be to split the PR up into multiple smaller ones, perhaps with some grouping of "related" changes. This would simplify the reviewing the changes (and whether everything was correctly adapted for Q2RTX), making things a bit safer, at the cost of taking longer.

Disclaimer: this is all just my personal opinion, I don't decide what gets merged or not.

Paril:
> This is quite a large change with 355 commits. This is hard to review ... e.g. I don't think I have the stamina to do this.
> 
> However, just from a quick look, it seems the changes range from maintenance without semantic changes (like qboolean -> bool) to e.g. breaking changes (savegame format) - there would probably be different considerations for merging these changes (or not). My recommendation would be to split the PR up into multiple smaller ones, perhaps with some grouping of "related" changes. This would simplify the reviewing the changes (and whether everything was correctly adapted for Q2RTX), making things a bit safer, at the cost of taking longer.
> 
> Disclaimer: this is all just my personal opinion, I don't decide what gets merged or not.

If I recall the only breaking change was the save format, because of the qboolean changes & float -> int frametimes. Everything else is maintenance that Q2PRO was doing along the way - various changes to error reporting, asynchronous screenshots, and a lot of simplification (hence the net negative in total code). The OpenGL backend was also overhauled quite heavily, but luckily since Q2RTX hadn't touched any of it it was (mostly) cleanly patched. The legacy SDL backend was removed, as well as some other systems that Skuller probably decided weren't being used at all. The most trouble I ran into (in terms of differences) was with screenshotting, since that entire image system uses stbi in RTX, but it was simple to adapt to Q2PRO's async queue.

The problem with splitting up the PR is that nearly all of these changes are chronologically required, or depend on each other in some way. It would take me a lot longer to figure out the relationship between all of the commits than it was to adapt these as-is. I don't know if I have the stamina to do *that*, heh.

I took care in ensuring nothing was lost in the transitions; a lot of the more important commits were already merged along the way before this, luckily. Commit 61c1eabc7de22b546131ad4e56dd975dd971f323 is probably the one that touches the most files, so if you're somehow able to diff the range excluding changes made by that commit it should be a bit easier to grok.

I'd wager maybe 70% of patches were able to apply with zero intervention on my end (the ones that show they're committed directly by the original authors were straight patches; the ones that begin with "From" are ones I had to do by hand, so those are the ones with the highest *potential* rate of error, while the rest are equivalent to having pulled it straight from the upstream repo.

apanteleev:
I briefly looked at each of these 350 commits... Generally these are engine/networking/GL changes, seem relatively safe assuming that they've been around in Q2PRO for a while. We still need to test the updated Q2RTX pretty heavily though.

A few notes:

e95c3cb6ede0c9960c27df1c79fadc15b14a3f54 - Disable system DPI scaling on Windows.

Similar change is already there in Q2RTX, should probably remove the old files.


86c093155314baffebce119a5b5ca63d189ca4e4 - Make light styles always compiled in.

Need to make sure it doesn't break light styles in Q2RTX.


7db5656ed47ad555b887ba92f26b3b161298bf36 - Parse userinfo ‘fov’ value as integer.

Why?


26b3007d6205afb8a562e67f48ca1938cabe9748 - properly detect git repository, works in submodule too
c03365054920ec8bf19c30b2be986f741c0c2098 - Simplify revision number calculation.

I don't think version.sh is used in q2rtx


8f15ac532843b791c97a82c747fb8046a28466f2 - Switch to the new QGL implementation.

Maybe we can remove the GL headers from the repo?


6c92593b95aaa673c1ed07bcbc4183de9fd0dc85 - Add experimental GLSL rendering backend.

Nice, this addresses #153
Also, GL headers.


e575a982082c3dd911c448882f7fc35f51bd7afa - Using separate type for error values only adds clutter. Replace with int.

I disagree with the subject statement. Returning qerror_t declares that it's a status code and not some other int.


cf3a9f90f929804a1ced42b7375aa85c29b1110c - Use SSE math by default for x86 builds.

These cmake changes should be conditional on x86/x64 target arch (not ARM).


2b60003ed7d4269f6dd1d07b07cecb6743695134 - Convert nextthink to frame numbers.
and a few more after it

Is this a good thing?...


23ef7b7e5aeda9b269069ee1c02515ae8ad9befe - Add previous weapon to key bindings menu.

q2pro.menu should be deleted

res2k:
> [2b60003](https://github.com/NVIDIA/Q2RTX/commit/2b60003ed7d4269f6dd1d07b07cecb6743695134) - Convert nextthink to frame numbers. and a few more after it
> 
> Is this a good thing?...

Using integer frame numbers should be more robust for long-running games, when you get into `float` value ranges with insufficient precision to represent a difference of 0.1. At least that's my guess to the reason behind this change.

Of course this assumes that the frame rate visible by the game is always 10 fps. Looking at `SV_Frame()`, this is the case.

res2k:
> [86c0931](https://github.com/NVIDIA/Q2RTX/commit/86c093155314baffebce119a5b5ca63d189ca4e4) - Make light styles always compiled in.
> 
> Need to make sure it doesn't break light styles in Q2RTX.

I looked at the commit. Unfortunately I noticed a change unrelated to lightstyles itself:
```diff
-   con_scale = Cvar_Get("con_scale", "1", 0);
+   con_scale = Cvar_Get("con_scale", "0", 0);
```
I noticed this by chance. Maybe it's the only instance of an unrelated change that slipped in.
Still, it isn't really that great if you can't rely on a commit doing what it's description says...

Paril:
@res2k the 86c0931 commit issue was my fault, not Skuller. See https://github.com/skullernet/q2pro/commit/dfdbce648061261808ccef2918f3d75c22740476

This change actually comes from https://github.com/skullernet/q2pro/commit/e5a22c4263b381c25c5525130cc8302f37eb3fae
(commit 5870f6d106d6774babe3bf7b4085868b3e6828fa in Q2RTX) but for whatever reason I accidentally shifted it into the next commit.

Like I said before, the ones that start with "From" are the ones that need additional scrutiny since they were ones I had to do by hand, so they are the most likely to contain issues like this. I appreciate the find! I'm not sure if I want to risk re-writing the Git history to fix this one though.

Paril:
Oh, I should also mention - I left all of the documentational changes out of these commits. I can do a final pass over the q2pro docs and change the q2rtx docs appropriately.

Paril:
> 23ef7b7 - Add previous weapon to key bindings menu.
> q2pro.menu should be deleted

Agreed, but having it be modified helped me be aware of any changes that were required; for instance choice menus use $$ instead of $ now. I did bring that over to q2rtx.menu.

Issue #167, opened 2021-12-13
Crash if fake emissive images are enabled and image has no diffuse map


Paril:
`get_fake_emissive_image` can crash if the map has a texture with an invalid diffuse map (missing file, for instance) and attempts to create a fake emissive image from it.

PR #168, opened 2021-12-14
Add a NULL check to vkpt_fake_emissive_texture. For #167

res2k:


Paril:
I did a static analysis run earlier, and found a few other potential null pointer dereferences. I fixed these in my local branch, but here's the outline:

`create_poly`, the initialization to emissive_factor could deref null if texinfo has no material. The fix:
```
	const float emissive_factor = (texinfo->c.flags & SURF_LIGHT) && texinfo->material && texinfo->material->bsp_radiance
		? (float)texinfo->radiance * cvar_pt_bsp_radiance_scale->value
		: 1.f;
```

`collect_sky_and_lava_light_polys`, the branching here is a bit confusing because there's a test for `surf->texinfo`'s existence, but the function then later uses it regardless of that outcome. I changed the loop to break entirely if texinfo is not set.

```

		if (belongs_to_model(bsp, surf))
			continue;
		if (!surf->texinfo)
			continue;

		int flags = surf->drawflags | surf->texinfo->c.flags;

		bool is_sky = !!(flags & SURF_SKY);
```

in `material_command`, the function assumes that `MAT_SynthesizeEmissive` will always set `image_emissive`, even though as discussed with your change it is not always the case. Adjustment was simply:
```
			// Regenerate emissive image
			MAT_SynthesizeEmissive(mat);
			// Make sure it's loaded by CL_PrepRefresh()
			if (mat->image_emissive) {
				IMG_Load(mat->image_emissive, mat->image_emissive->pix_data);
			}
			reload_flags |= RELOAD_MAP;
```

I did these in a larger commit so I can't easily format a patch. If you absolutely need it in patch format I can whip one up later.

apanteleev:
Merged in bafa1dcf67641bcdd86033b16f2b4f7853ab4a3a
Thanks!

PR #169, opened 2021-12-15
move cameras to per-map files

Paril:
This PR moves cameras away from the single-file system, which is impossible for modders to use for custom maps (with the current system it's impossible to allow downloads of camera files without overwriting all of their cameras, for instance).

One potential improvement is to also allow embedding camera entities in a map, which are ignored by the game code and solely used by the renderer, which eliminates an extra file for new maps, but that's much lower priority.


baseq2/maps/cameras/lab.txt
```diff
@@ -0,0 +1 @@
+(-2008.9, 678.2, 206.1) (0.001, -0.810, -0.586)
```
res2k, line 1: the to original `cameras.txt` had a comment for the "lab" map.

src/refresh/vkpt/cameras.c
```diff
@@ -0,0 +1,86 @@
...
+		vec3_t pos, dir;
+
+		if (sscanf(linebuf, "(%f, %f, %f) (%f, %f, %f)", &pos[0], &pos[1], &pos[2], &dir[0], &dir[1], &dir[2]) == 6)
+		{
+			if (wm->num_cameras < MAX_CAMERAS)
+			{
+				VectorCopy(pos, wm->cameras[wm->num_cameras].pos);
+				VectorCopy(dir, wm->cameras[wm->num_cameras].dir);
+				wm->num_cameras++;
+			}
```
res2k, line 58: Could be helpful for map authors to print a message if `MAX_CAMERAS` is exceeded.

baseq2/maps/cameras/lab.txt
Paril: oops, thanks

Paril:
Pushed with adjustments requested by @res2k

PR #171, opened 2021-12-18
Primitive storage and static model improvements

apanteleev:
This PR may look a bit odd coming from me, but I just want to have some more eyes on the changes before I merge them... It's large. The changes will be merged with squash most likely.
@res2k @Paril specifically

Summary:
- Replaced the arrays of primitive attributes that were specific to BSP or models with a single, unified storage format: VboPrimitive.
- Added support for static models with their own BLAS that never go through the instnace_geometry shader. This significantly improves performance when models become more complex, think hundreds of thousands of triangles.
- Increased the limits for primitive and mesh counts, removed the limits on BSP or instanced buffer sizes.
- Generally simpler and more logical code when it comes to primitive handling.


src/refresh/vkpt/shader/vertex_buffer.h
```diff
@@ -42,8 +41,8 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 
 #define ALIGN_SIZE_4(x, n)  ((x * n + 3) & (~3))
 
-#define BSP_VERTEX_BUFFER_BINDING_IDX 0
-#define MODEL_DYNAMIC_VERTEX_BUFFER_BINDING_IDX 1
+#define PRIMITIVE_BUFFER_BINDING_IDX 0
+#define POSITION_BUFFER_BINIDNG_IDX 1
```
res2k, line 42: Typo ("BIN _ID_ NG" instead of "BIN _DI_ NG")

src/refresh/vkpt/vertex_buffer.c
```diff
@@ -23,58 +23,395 @@ with this program; if not, write to the Free Software Foundation, Inc.,
...
+	for (int i = 0; i < bsp_mesh->num_models; i++)
+	{
+		bsp_model_t* model = bsp_mesh->models + i;
+		create_model_blas(&model->geometry, qvk.buf_world.buffer);
+	}
+
+	uint8_t* staging_data = buffer_map(&staging_buffer);
+	memcpy(staging_data, bsp_mesh->primitives, bsp_mesh->num_primitives * sizeof(VboPrimitive));
+
+	mat3* positions = (mat3*)(staging_data + bsp_mesh->vertex_data_offset);  // NOLINT(clang-diagnostic-cast-align)
```
res2k, line 302: IMO, since this is dealing with triangles, using a type (name) other than `mat3` could be clearer.

src/refresh/vkpt/vertex_buffer.c
```diff
@@ -50,14 +50,54 @@ static BufferResource_t null_buffer;
 // https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/VkAccelerationStructureCreateInfoKHR.html
 #define ACCEL_STRUCT_ALIGNMENT 256
 
+void vkpt_init_model_geoetry(model_geometry_t* info, uint32_t max_geometries)
```
res2k, line 51: Typo: Missing 'm' in geometry

src/refresh/vkpt/vertex_buffer.c
```diff
@@ -1044,9 +1054,73 @@ vkpt_vertex_buffer_upload_models()
...
+		return;
+	
+	vkDeviceWaitIdle(qvk.device);
+
+	destroy_primbuf();
+
+	prim_count = (uint32_t)align(prim_count, PRIMBUF_SIZE_MIN);
+	Cvar_SetInteger(cvar_pt_primbuf, (int)prim_count, FROM_CODE);
+
+	Com_WPrintf("Resizing the animation buffers to fit all meshes. Set pt_primbuf to at least %d to avoid this.\n", prim_count);
```
res2k, line 1156: Hmm... Wonder if that perhaps should be a "developer" printf?
Or, since `pt_primbuf` is archived, is this necessary at all? The new (larger) value will be stored in the config anyway...

res2k:
I've found some minor things (typos etc), but there's nothing major (...that I noticed).

apanteleev:
Thanks for the review! All feedback has been addressed.

Paril:
Somehow I missed this PR entirely in my inbox - I'm sorry! I'll go over the commits even though it's too late, haha

src/refresh/model_iqm.c
```diff
@@ -482,30 +482,30 @@ int MOD_LoadIQM_Base(model_t* model, const void* rawdata, size_t length, const c
...
 		}
 	}
 
 	if (header->num_joints)
 	{
-		CHECK(iqmData->jointNames = (char*)MOD_Malloc(joint_names));
-		CHECK(iqmData->jointParents = (int*)MOD_Malloc(header->num_joints * sizeof(int)));
-		CHECK(iqmData->bindJoints = (float*)MOD_Malloc(header->num_joints * 12 * sizeof(float))); // bind joint matricies
-		CHECK(iqmData->invBindJoints = (float*)MOD_Malloc(header->num_joints * 12 * sizeof(float))); // inverse bind joint matricies
+		CHECK(iqmData->jointNames = MOD_Malloc(joint_names));
```
Paril, line 491: this just reminded me, on my local branch I changed jointNames to be an array of `char[MAX_QPATH]` so that I can easily index the joint names at runtime. Currently, you need to loop over the string via strlen which is pretty inefficient.. not that it totally matters I guess since you shouldn't be looping through joint names every frame at runtime, but just a random note that I think a single long string of "name data" is a bit harder to parse through than just fixed-length string indices.

PR #173, opened 2021-12-29
separate out nearest filtering for UI, default it on

Paril:
Split pt_nearest into two variables (the second one is named pt_nearest_2d), and default it to on, to match other Q2 engines/not cause blurry low quality UI textures.

Also fixes blurry non-standard fonts, since it was hardcoded to blur only the main font.

### DIFF
--- a/src/refresh/vkpt/main.c
+++ b/src/refresh/vkpt/main.c
@@ -68,7 +68,8 @@ cvar_t *cvar_pt_projection = NULL;
 cvar_t *cvar_pt_dof = NULL;
 cvar_t* cvar_pt_freecam = NULL;
 cvar_t *cvar_pt_nearest = NULL;
-cvar_t *cvar_pt_nearest_2d = NULL;
+cvar_t *cvar_pt_bilerp_chars = NULL;
+cvar_t *cvar_pt_bilerp_pics = NULL;
 cvar_t *cvar_drs_enable = NULL;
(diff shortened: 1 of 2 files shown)


res2k:
FWIW, the GL renderer seems to have two different cvars for this (`gl_bilerp_chars`, `gl_bilerp_pics`), maybe worthwhile to copy that approach? (The two vars also have different defaults - `..._chars` defaults to off, `..._pics` defaults to on.)

And/or, since the setting mostly concerns the UI, why not put "UI" somewhere in the cvar name, e.g. `pt_nearest_ui` or `ui_nearest`?

Paril:
Not a bad idea. We could just copy the GL ones entirely, then it'd be closer parity.

Paril:
Changes done, now it's mostly parity with the GL system apart from the interpretation of the 1 value for pt_bilinear_pics.

apanteleev:
Why not just reuse the same `gl_bilerp...` vars?

Paril:
I guess you could - it's usually standard in engines with different renderers to use different cvars, since the `gl` ones don't make sense for a non-GL renderer. There's also the issue of 1 being semantically different in GL, since the VK renderer doesn't have the "scratch" system, so we'd have to make 1 effectively the same as 0 which might be a bit confusing (2 would be "enabled" for VK).

Issue #185, opened 2022-02-13
Animated textures don't work in 1.6


Paril:
I was informed that animated textures seem to be broken in 1.6; buttons, for instance, don't animate. We haven't checked func_wall's or non-bmodel animated textures.

res2k:
Looks like they fell by the wayside in #171.
Their "animation" (really, choosing material frame based on entity) was handled by the `instance_geometry` shader, but now that's not used any more for brush models...

PalmliX:
Just confirmed, created a simple test map, in 1.5 animated textures on a func_wall work correctly, in 1.6 there is no animation. Animated texture on worldspawn works the same in both maps. I uploaded the test map if it helps.
[texanim.zip](https://github.com/NVIDIA/Q2RTX/files/8063147/texanim.zip)

apanteleev:
Thank you all for reporting, investigating, and fixing the issue.
I merged the PR in https://github.com/NVIDIA/Q2RTX/commit/73c0b7f202d9a2a88fecf787723f34a50f5a0031

PR #187, opened 2022-02-14
Fix bsp model texture animation

res2k:
Fix missing texture animation on bsp models (issue #185) by "animating" the frame in `load_and_transform_triangle()`.

Another approach I considered was to go over each bsp model once per frame and animating the material on each primitive - think similar principle as `instance_geometry`, only for bsp models, and material animation only.

I went for the `load_and_transform_triangle()` approach mainly because it's simpler.
However, it may have a performance disadvantage, as the "animation" would potentially be done quite a number of times, compared for guaranteed once per frame in the approach above.
(Perhaps it would be worthwhile to implement that as well and check if there's a measurable performance difference.)

### DIFF
--- a/src/refresh/vkpt/shader/instance_geometry.comp
+++ b/src/refresh/vkpt/shader/instance_geometry.comp
@@ -206,21 +206,8 @@ main()
 		if (mi.material != 0)
 			t.material_id = mi.material;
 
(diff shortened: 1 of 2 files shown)


PalmliX:
Thanks for this res2k! I'm experimenting with some very long animations, over 100 frames, so I'll let you know about the performance.

Issue #178, opened 2022-01-22
Saved games doesn't work after update to 1.6.0


mistificator:
I've updated from 1.5.0 to 1.6.0 yesterday (I use Steam), and got these messages.
![image](https://user-images.githubusercontent.com/2280081/150647621-a2ef8e62-8f67-4db9-b544-7170cec4af04.png)

mistificator:
Workaround is to take old gamex86_64.dll from 1.5.0.

Paril:
You can use the old game DLL, but this is to be expected with Q2-based games. Save games will break between major game DLL version changes.

There were some legacy bugs fixed between the two, hence the version change.

mistificator:
So you think that in-game progress is nothing. Okay I see your point.

Paril:
> So you think that in-game progress is nothing. Okay I see your point.

This is just how it was in Quake II. You're free to contact John Carmack and let him know you don't like his 15+ year old decisions. We did inherit some pointer-esque fixes from Q2PRO but because the save format is extremely basic and basically just writes the raw memory to disk, this is bound to happen. It's out of the scope of Q2RTX to introduce a new save system, IMO. You should file an issue with Q2PRO instead; if they adopt one, Q2RTX will inherit it.

res2k:
@mistificator I attempted to add savegame compatibility for 1.5.0 in PR #193. It seems to work, but I gave it only a quick test, so maybe give it a try. You should be able to open savegames from 1.5.0, and after re-saving they'll be in the updated format. (Note this doesn't work in the other direction: savegames created with 1.6.0 will never work in 1.5.0.)

apanteleev:
PR #193 is now merged, which should take care of this issue.

PR #193, opened 2022-02-25
Add save game compatibility with Q2RTX 1.5.0

res2k:
Luckily, the general structure is not _that_ different, so it's possible to load old (1.5.0) savegames by doing some things slightly different when the old version is encountered.

* Use a special field type for 'frametimes', covert to framenum on load
* Store the old pointer table as well

Addresses #178.


res2k:
> we'd still have to have everybody lose progress just to add that support for forwards-compatibility, heh.

FWIW, my first idea to provide savegame compatibility with 1.5.0 was to "let's copy the savegame logic wholesale", until I realized the existing code could be tweaked with the checks/conditions. But that approach would of course work if any other savegame changes were to be made.

I'm not sure some "custom" savegame system would be worth it. Perhaps if it was inherited from Q2PRO... can't say what they're thinking, but the past suggests a bit that savegame compatibility may not be a priority for them.

res2k:
> Honestly, it might just not be worth it. Q2 includes several config files in pak0.pak for starting the game at a certain unit, and there's no "progress" stored between units anyways... if mission pack support was in the cards I'd potentially be on board but this is _just_ baseq2, so, I dunno.

Well, _some_ progress is stored between units - e.g. items, weapon, ammo. See `WriteGame()` in baseq2 for details.
Since most of the save game logic lives in the game DLL the mission packs would be unaffected by the recent savegame changes (as they didn't touch the parts in the server).

FWIW, at least from a "user experience" perspective savegame compatibility is welcome and worth it. Even if you can, in theory, cheat your way to the state you had before, it's still a feeling of having "lost" investment into the game. (Also, not everybody may know how to actually cheat yourself to a certain place, or may not want to.)

Paril:
Of course, but it's trivial to restore - and like I said the unit configs will get you at the start of a unit with the weapons you can get from the prior. I know it's not the same, just thinking ahead to this happening again down the road.

And yes, I know the mission packs are unaffected since they are different mods, but my point was that if we were updating all three game DLL versions at the same time maybe there was more of a justification for it.

PR #194, opened 2022-02-26
Implement flare on top of actual game

res2k:
This is a bit more unusual one.

Adds flare support inside a "proxy game" that layers on top of the actual game.
The motivation is to support mods, games for which the source code may not be available, but which nevertheless suffer the Q2RTX "problem" of occasional darkness due to lack of light-emitting surfaces.

Moving the flare logic out of the principal game changes it's behavior somewhat; most significant is probably that it's not a selectable weapon any more (throwing a flare is done via a command, `throwflare`) and does not consume ammo (grenades) any more (which is, IMO, not necessarily a bad thing).

Also thrown in at the end is a bit of logic to improve compatibility between RTX and non-RTX clients.


apanteleev:
While I understand the motivation behind this pull request, I think this is just too hacky. A basic flashlight would be a much cleaner solution to the problem.

res2k:
Hmm, I think I thought about this as well... actually that was some time ago now.
IIRC my "cons" of wrt to such a flashlight were:
(a) Technical limitations - in the sense that a proper flashlight would best be implemented using a spotlight, which the renderer (currently) lacks. (It's also possible to "emulate" a flashlight using a cleverly placed point light, but I fear that may look a bit meh.)
(b) Interestingness - a flare is basically a temporary, user-placeable light, which can be dropped at some point, yet you can move around and see how it affects lighting/shadowing on varying parts of the world and items, showing off that gorgeous raytracing ;)

But in general, something purely client sided would probably the cleanest to implement... but doing something "physical", like the current flare, is rather challenging to do client-side only, hence that funky "overlay" game.
But something "non-physical",  like a flashlight, should probably be quite simple to do client-wise.

What would be your preferred approach for a flashlight? A proper spotlight, a "moving point light", some other approach?

apanteleev:
A proper spotlight, maybe even a textured one. It's not that difficult to extend the light sampling code to add a spotlight term to the sphere lights, and throw in some basic RIS on the sphere lights to improve quality. I can take a look at that, not sure when though.

res2k:
Sorry, could you unabbreviate "RIS"?

apanteleev:
RIS = Resampled Importance Sampling, aka Importance Resampling.
When you take a few random items from a pool with a uniform PDF, and then select one of them based on a different PDF. Same thing already happens for triangle lights.

PR #201, opened 2022-03-06
Ignore a material definition if a game overrides a texture

res2k:
Some games override baseq2 assets without changing the name -
e.g. 'action' replaces models/weapons/v_blast with something looking completely differently.
Using the material definition from baseq2 makes things look wrong.
So try to detect if the game is overriding a texture from baseq2 and, if that is the case, ignore the material definition (if it's from baseq2 - to allow for a game-specific material definition).

(I previously submitted something with the same intent as part of #116 [[2415da2](https://github.com/NVIDIA/Q2RTX/pull/116/commits/2415da21620e07aeada906780097512ac7689efd)], but it seems those changes got lost or ineffective during the material system overhaul.)

### DIFF
--- a/src/refresh/images.c
+++ b/src/refresh/images.c
@@ -999,24 +999,6 @@ static int _try_image_format(imageformat_t fmt, image_t *image, int try_src, byt
     if (!data) {
         return len;
     }
-    /* Don't prefer game image if it's identical to the base version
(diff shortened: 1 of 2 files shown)


apanteleev:
Is it expected that `base1` looks like this after these changes? ;)

quake001

res2k:
Oops 🙈

Should work with after the last commit...

PR #203, opened 2022-03-11
Spotlight/Flashlight (WIP)

res2k:
Following the discussion in #194 here the (start of) an attempt to implement a client-side flashlight.
It doesn't contain any of the "fancier" stuff that was mentioned (texturing or "RIS on the sphere lights to improve quality") but I think it's good enough to collect some feedback on the implementation.

To enable it set `cl_flashlight` to 1. Works really well with a toggle bind, e.g. `bind f "toggle cl_flashlight"`.

In RTX mode it's a proper spotlight.
In GL mode, to have a bit of feature parity, I added a fake flashlight - ie placing a point light on the surface the player currently points at. It's also a bit crappy as it completely ignores entities when determining the light position. TBH, I'm actually not sure that fake flashlight is worth the effort.

Screenshot (made in one of the dark parts in `strike`):
image


res2k:
Hey @apanteleev , did you have for a look at this?
While there's more to be done (see description), what do you think about the general approach?
And perhaps warranting a closer look would be the changes to `sample_dynamic_lights`/ex-`sample_spherical_lights`.
Thanks!

apanteleev:
Sorry, I didn't have a look yet. I'll try to do it today or tomorrow.

apanteleev:
Tested the changes. Looks pretty good, though I do have some suggestions:
- I think the intensity of 10000 looks better than the default 2000;
- Perhaps move the emitter away from the camera, to the bottom-left or bottom-right, so that it would look like a flashlight in hand, and not glowing eyes ;)
- Would be nice to have a more realistic flashlight emission profile, rather than an ideal spotlight. Something with a pronounced central lobe and a dim/wide falloff. Like (4) or (12) on this picture: https://www.blender3darchitect.com/wp-content/uploads/2019/01/ies-lights_700px_op.jpg This could be a later improvement though.

The code looks good.

res2k:
> Tested the changes. Looks pretty good, though I do have some suggestions:
> 
>     * I think the intensity of 10000 looks better than the default 2000;

Sure, can change that.
 
>     * Perhaps move the emitter away from the camera, to the bottom-left or bottom-right, so that it would look like a flashlight in hand, and not glowing eyes ;)

Ok. I'll take a look at some muzzle flash locations, that should look similar to a "handheld" flashlight.

>     * Would be nice to have a more realistic flashlight emission profile, rather than an ideal spotlight. Something with a pronounced central lobe and a dim/wide falloff. Like (4) or (12) on this picture: https://www.blender3darchitect.com/wp-content/uploads/2019/01/ies-lights_700px_op.jpg This could be a later improvement though.

Maybe something that could be approximated with a projective texture, though the examples you picked certainly look more complex. (Might be worthwhile to search around a bit, see what kind of data Blender consumes to produce such lights.)

> The code looks good.

Great! Thanks for the review.

Do you have an opinion on the "fake" GL flashlight?
The effect is a bit crappy now and I'm not sure it's worth the effort to make it look good,
but if you think it's a valuable addition I'll look into improving the interaction with entities a little.

Beilerbey:
> Maybe something that could be approximated with a projective texture, though the examples you picked certainly look more complex. (Might be worthwhile to search around a bit, see what kind of data Blender consumes to produce such lights.)

Those are IES files, you can find out more here (not sure if they would work in Q2RTX though): https://ieslibrary.com/en/home

res2k:
> Those are IES files, you can find out more here (not sure if they would work in Q2RTX though): https://ieslibrary.com/en/home

No, they won't work as-is... Nevertheless, thanks for the pointer, I'll look into it!

apanteleev:
> Do you have an opinion on the "fake" GL flashlight?

Well... placing a sphere light at the nearest wall seems like a reasonable solution. But it could be tuned better: increase the tracing range; increase the intensity/radius when the wall gets further away. I'm not sure anyone would care for a flashlight in the GL version though.

> Those are IES files

Correct. IES files are fairly easy to parse, too; I wrote a custom parser for them sometime last year, here it is:
https://github.com/NVIDIAGameWorks/donut/blob/main/src/engine/IesProfile.cpp
https://github.com/NVIDIAGameWorks/donut/blob/main/shaders/ies_profile_cs.hlsl

But I don't think we should use that in Q2RTX. It could be either a simple 1D texture loaded from file, that maps the angle from the axis to an intensity, or an analytic approximation of that texture. A reasonably simple approximation could be cheaper at render time, too.

Beilerbey:
Yes I had a feeling they wouldn't work. I do have a texture that could be suitable and/or I could help make one without problems, I know nothing on a technical level (I'm an artist not a programmer) so someone would've to tell me specifics (format, size, bit-depth, etc), but other than that I'd be happy to help. Here is what I had already made, in blender it looks like this (notice that the render is noise free, the little dark spots are in the texture and are supposed to be dust/dirt on the glass): 
Spotlight

res2k:
I have made a bunch of adjustments:
* Upped default intensity.
* I opted to remove the GL flashlight. However, since it's still in the git history, the implementation could be resurrected later, if so desired.
* Tweaked the position so it's positioned more as if it was a gun attachment.

apanteleev:
Looks pretty good now, thanks for the update!
What are your thoughts on implementing an emission profile? Later?

res2k:
> Looks pretty good now, thanks for the update! What are your thoughts on implementing an emission profile? Later?

Yes, later :)

But while I'm here let me share my current thoughts/plan.
On the engine side I like the "a simple 1D texture loaded from file, that maps the angle from the axis to an intensity" approach.
Using a texture allows a bit more flexibility than an analytic approximation - probably something of interest for modders and forkers (or perhaps Q2RTX itself, should a spotlight be something suitable for more effects).

As to getting such a texture: ieslibrary.com seems to have profiles that are axially symmetric, those would be a perfect fit for that approach. (Unfortunately you can't search for those profiles, so you can only pick a nice-looking profile and hope it's symmetric...)

Of course, such an 1D texture needs to be created; currently I'm writing a small Python script to do so, to have something going. But eventually it's probably sensible/helpful to have such a script, or tool in the Q2RTX repo... Do you agree? If so, do you have a preference on how to approach this - I'm thinking a Python script should be fine, but maybe you prefer something in C or C++.

apanteleev:
A Python script would be fine for a tool like that.

The MR is good to merge now, right?

res2k:
> The MR is good to merge now, right?

From my side, yes.

PR #213, opened 2022-04-21
Move identical image detection to filesystem

res2k:
Previously, a check was added to reject "game" images if it was identical to the "base" image, to allow image overrides to be used in that case (commit cf65b63b5d3a1927565e89933c18cc23cb408811).
However, this clashes a bit with #201, as this detects a texture overriden by the game and so doesn't use the material definition from base. However, at the same time, loading the image from the game will be rejected as it's identical to the base image, leading to no image at all being loaded...

Both aspects are desireable - if a game just copies an image, it should use any high-res override and the material definition from baseq2, but if it's a different image, neither override nor existing material definition should be used.

I opted to approach the issue on a different level - the filesystem. Basically, if we find a file in a .pak that already exists in baseq2, we do the identity check there, and reject files in game .paks that are identical to the baseq2 version.
That way, the material system will never see any duplicated images and just use the baseq2 version and any override and material definition from there.

### DIFF
--- a/src/common/files.c
+++ b/src/common/files.c
@@ -2062,7 +2062,7 @@ static void pack_hash_file(pack_t *pack, packfile_t *file)
 
 // Loads the header and directory, adding the files at the beginning
 // of the list so they override previous pack files.
-static pack_t *load_pak_file(const char *packfile)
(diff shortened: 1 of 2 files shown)


res2k:
Abandoned due to second thoughts. Added a different solution to #201.

PR #214, opened 2022-04-24
Add emission profile to flashlight

res2k:
Follow-up to #203. Now uses an emission profile generated from an IES file.

Screenshot from master:
quake000

Screenshot with emission profile:
quake001

* I tried to extend the light sampling in a way that allows future enhancements, e.g. perhaps profiles that are not axially symmetric, or projection lights.
* I opted to store the emission profile as a 16bpc grayscale PNG, so as a collateral I needed to add support for that.
* The emission profile texture is indexed by the cosine of the angle from the spotlight direction. It was quite convenient to simply use that.
* The specific used profile is BEGA 50964K4 from ieslibrary.com

Possibly future enhancements:
Currently, the emission profile is assumed to cover 90 degrees from the spotlight direction.
Might be interesting to make that customizable, e.g. for profiles that just zero out before 90 degrees, or for profiles that cover a larger range (the IES profile mentioned above actually has values for a tad over 90 degrees).


apanteleev:
Thanks @res2k - this looks good overall, but I'd like to request one change. Can you make the profile texture have `angle` on the X axis and not `cos(angle)`? With the cosine, you get really poor angular resolution next to the center of the beam.

res2k:
> Can you make the profile texture have `angle` on the X axis and not `cos(angle)`? With the cosine, you get really poor angular resolution next to the center of the beam.

Sure, updated the branch.

Issue #170, opened 2021-12-16
Use OpenAL Soft


Paril:
https://github.com/kcat/openal-soft

I'm unsure if Mac & Unix require it (they might already use oal-soft internally), but Windows will definitely need this. Creative's final OpenAL redistributable has a lot of bugs that make it unsuitable as a drop-in sound replacement - for instance, sounds played have a built-in, unavoidable "fade-in" effect (most noticeable on blaster & machinegun).

OpenAL-Soft is LGPL compatible and fixes these bugs. Q2pro recommends it for Windows as well, as it makes the game sound indistinguishable from Q2's old software drivers.

res2k:
@Paril, I added a PR that provides the OpenAL-soft DLLs on Windows.

I didn't notice too much difference to the "default" OpenAL implementation used on my system, but maybe I'm too much of an audio cretin for that, or that implementation was newer/different from Creative's final OpenAL redistributable.
But if you experienced issues with OpenAL in the past perhaps leave a comment how this worked out for you.

(FWIW, I believe OpenAL on Linux is typically openal-soft anyway. Don't know about Mac.)

Paril:
Cool! Yeah, creatives oal driver has issues with sounds having weird delay at the beginning (like a fade in), some sounds not playing at all (Adam ran into this with our fork with some looping sounds not working on créatives driver) and of course it lacks all of the extensions (loop points and efx are great). In NAC I ended up just pulling in oal in cmake and embedding it statically so people don't have to worry about DLL hell.

Paril:
If you're interested in that process, although I dunno if Q2RTX wants to go the static route, the commits of interest are here:
https://github.com/Paril/Q2RTX/commit/81814f6443602ae6dd6685f8623bb4db5a074b11
https://github.com/Paril/Q2RTX/commit/ecf85504c3e1b907e0a470b3014ea29e7ce30f08 (cmakelists changes only)

it's nice to not have to worry about peoples' conflicting DLLs, but I don't know if this builds on Unix/Mac atm and people wouldn't be able to swap out their DLLs with newer ones if they had fixes. Give & take I suppose

apanteleev:
Hi @Paril - could you make a PR for the static oal-soft please? I'd rather merge that than the DLL version.

Paril:
Hey!
Yeah I can. The only thing is I couldn't test the Unix/Mac build processes.

res2k:
[...]

PR #215, opened 2022-04-24
Provide OpenAL soft

res2k:
On Windows, drop OpenAL soft DLLs next to the client. For #170


apanteleev:
Merged https://github.com/NVIDIA/Q2RTX/pull/224 instead, closing. Thanks anyway.

PR #221, opened 2022-05-07
Wayland support

### LINKED ISSUE #217: Q2RTX still uses Xwayland on Wayland
Hello All!

I found that with the latest version 1.6.0 of Q2RTX still uses XWayland (X.Org) even if I use 100% Wayland session. 
I use Manjaro Linux with Gnome GUI and AMD Vega 56 GPU. 

If I force SDL_VIDEODRIVER=wayland then the game crashes on the startup. 

Is it possible to run the game without XWayland ?

Thanks!

res2k:
Adds a bunch of fixes to make Q2RTX run on Wayland. Hopefully fixes #217.

Note that this contains some changes (dcd28504cd3b122e77962ae2152e41a8858d4f84, 269ad3cec304ac230d88f48988b9e7670d69ba6b) that may actually be issues somewhere in the software stack I used (Wayland on NVIDIA) - they may not be necessary for Wayland on AMD, but I can't test that.

### DIFF
--- a/src/refresh/vkpt/draw.c
+++ b/src/refresh/vkpt/draw.c
@@ -524,8 +524,8 @@ vkpt_draw_create_pipelines()
 		.srcColorBlendFactor = VK_BLEND_FACTOR_SRC_ALPHA,
 		.dstColorBlendFactor = VK_BLEND_FACTOR_ONE_MINUS_SRC_ALPHA,
 		.colorBlendOp        = VK_BLEND_OP_ADD,
-		.srcAlphaBlendFactor = VK_BLEND_FACTOR_ONE,
-		.dstAlphaBlendFactor = VK_BLEND_FACTOR_ZERO,
+		.srcAlphaBlendFactor = VK_BLEND_FACTOR_SRC_ALPHA,
+		.dstAlphaBlendFactor = VK_BLEND_FACTOR_ONE_MINUS_SRC_ALPHA,
 		.alphaBlendOp        = VK_BLEND_OP_ADD,
 	};
 
--- a/src/unix/video.c
+++ b/src/unix/video.c
@@ -459,6 +459,13 @@ bool VID_Init(graphics_api_t api)
     }
 
     VID_SetMode();
+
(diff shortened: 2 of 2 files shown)


apanteleev:
I have a couple questions/requests here:
- How come you need to install new Wayland-related packages for CI builds, but there are no build config changes at all, e.g. CMake settings? Does the newer SDL just require those libs?
- I don't think the `_VK(...)` change belongs here, or maybe it should be used only for RelWithDebInfo type builds.

res2k:
> * How come you need to install new Wayland-related packages for CI builds, but there are no build config changes at all, e.g. CMake settings? Does the newer SDL just require those libs?

Yes, SDL "wants" to enable Wayland support by default ("SDL_WAYLAND _(Wanted: ON)_" configure message), but only actually builds it if the necessary libs are present. So installing them is sufficient.

> * I don't think the `_VK(...)` change belongs here, or maybe it should be used only for RelWithDebInfo type builds.

Well, it was mainly added to possibly aid in troubleshooting for #217, though as it turned out, it didn't matter.
Though, I'm thinking getting Vulkan errors (somehow) reported in Release builds can be a troubleshooting aid...
perhaps consider making that check/reporting conditional somehow?

Last but not least, there's still the issue that the https://github.com/NVIDIA/Q2RTX/commit/dcd28504cd3b122e77962ae2152e41a8858d4f84, https://github.com/NVIDIA/Q2RTX/commit/269ad3cec304ac230d88f48988b9e7670d69ba6b where necessary to make things work on Wayland for me, but they don't feel entirely right to me...

res2k:
Rebased onto current master.

One of the "doesn't feel right" things (UNORM surface format) was done properly with #261.

That `VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR` seems to be ignored was also noticed by other people (https://forums.developer.nvidia.com/t/wayland-vulkan-ignores-vkcompositealphaflagskhr/202227/2).
This: https://github.com/NVIDIA/egl-wayland/pull/71#issuecomment-1310751909 suggests that some related change may come to a future driver update, so it might be worth to wait and see what that does.

PR #222, opened 2022-05-08
Store a material's "original" texture dimensions in special members

res2k:
…instead of the image_base width & height.

This allows to re-use the same "base" texture for multiple materials, even if
the original textures have different dimensions/aspect ratios.

Previously, when doing so, the image's width/height members would be repeatedly
overwritten, and whatever original texture was loaded last, that dimensions
would "win". When the various original textures had different aspect ratios
some texture scaling factors would've been wrong (and the texture appear
distorted).

(This doesn't occur with Q2RTX materials, but I stumbled upon that when creating some custom materials.)

### DIFF
--- a/src/refresh/vkpt/bsp_mesh.c
+++ b/src/refresh/vkpt/bsp_mesh.c
@@ -169,13 +169,9 @@ create_poly(
 	assert(surf->numsurfedges < max_vertices);
 	
 	float sc[2] = { 1.f, 1.f };
-	if (texinfo->material)
-	{
-		image_t* image_diffuse = texinfo->material->image_base;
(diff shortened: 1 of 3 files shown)


apanteleev:
Somehow, these changes break all polygon lights, they no longer emit anything.

res2k:
> Somehow, these changes break all polygon lights, they no longer emit anything.

Well, not _all_ of them, just those that aren't "entire texture emissive" 😬
(Of course, the map I noticed the size bug in, which has multiple poly lights, and which used to check the changes in, were _all_ "entire texture emissive". 🤦🏻‍♂️)

Issue #226, opened 2022-05-12
Permanently noisy world after certain visual effects


Paril:
When certain temporary visual effects are used, sometimes the result is that that area of the world becomes permanently noisy.

Initially I was able to repro this with a material on a TE_LIGHTNING bolt, but I couldn't get this to happen again with the latest versions. However, I found that RF_BEAM entities do repro the issue. Some mods use RF_BEAMs as replacements for blaster bolts, for instance, but I don't know if they're able to reproduce this on a large scale.

A way to reproduce this is to shim this code above the call to `prepare_entities` on line 2905:

```
	// TEST
	// debug bones
	int x = Cvar_Get("pt_test_debug", "0", 0)->integer;
	for (size_t i = 0; i < x; i++) {
		vec3_t r = { 40 + (crand() * 40), crand() * 40, crand() * 40 };
		particle_t *pr = &fd->particles[fd->num_particles++];
		memset(pr, 0, sizeof(*pr));
		pr->color = -1;
		pr->brightness = 1;
		pr->radius = 1.0f;
		pr->rgba = (color_t) { .u8 = { 255, 0, 0, 255 } };
		pr->alpha = 1.f;
		VectorAdd(cl.playerEntityOrigin, r, pr->origin);

		entity_t *b = &fd->entities[fd->num_entities++];
		memset(b, 0, sizeof(*b));
		b->skinnum = -1;
		b->rgba = (color_t) { .u8 = { 0, 0, 255, 255 } };
		b->alpha = 0.30f;
		b->flags = RF_TRANSLUCENT | RF_BEAM;
		b->frame = 8;
		VectorCopy(pr->origin, b->origin);
		vec3_t r2 = { 40 + (crand() * 40), crand() * 40, crand() * 40 };
		VectorAdd(pr->origin, r2, b->oldorigin);
	}
	// TEST
```

this produces a fake but similar output to what I'm doing with my bone debugger. Load up q2dm1, set `pt_test_debug` to a number (40 is what I was testing with, but various numbers will work), then set it back to 0. Notice now that the area you are in is permanently noisy even though there are no more lasers lighting the scene.

res2k:
> on line 2905:

Not sure this number is correct on current master; can you provide some context (lines before/after)?

Paril:
Directly before the call to "prepare_entities". There's only one call to it, so it should be easy to find. It's 2828 on master I think?

[...]

PR #227, opened 2022-05-12
Use exact number of cluster lights when assembling light lists in inject_model_lights()

res2k:
Previously, space was made in the light list offsets that amounted to the maximum
number of lights in a cluster, seen over the lifetime of the map. Unfortunately,
this affect light sampling after a large spike of dynamic lights in a cluster,
possibly resulting in a more 'noisy' appearance - as described in #226.

However, the exact number of lights in a cluster was actually already counted
and can be used to tightly pack the light lists, avoiding the permanent effect
on sampling after a spike of dynamic lights.

### DIFF
--- a/src/refresh/vkpt/shader/light_lists.h
+++ b/src/refresh/vkpt/shader/light_lists.h
@@ -172,12 +172,6 @@ sample_polygonal_lights(
 		
 		uint current_idx = light_buffer.light_list_lights[n_idx];
 
-		if(current_idx == ~0u)
-		{
-			light_masses[i] = 0;
-			continue;
-		}
-
 		LightPolygon light = get_light_polygon(current_idx);
 
 		float m = projected_tri_area(light.positions, p, n, V, phong_exp, phong_scale, phong_weight);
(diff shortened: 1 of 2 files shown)


res2k:
Indeed it's not that simple... 🙈

However, I had some ideas I could try:
1. `max_cluster_model_lights` is currently kept until the map is reloaded. But I think the maximum for a cluster would only have to be kept for the time it's relevant for rendering - that is, as soon a cluster leaves the PVS2 set for the current viewpoint, the maximum could be reset.
2. However, certain areas, like the aforementioned `security` with it's regularly flashing lasers, would be benefit from _not_ resetting the `max_cluster_model_lights` value to zero... so perhaps a bit of history could be kept of that number - like, a few seconds worth of client frames - and when the `max_cluster_model_lights` value would be reset as suggested in point 1, it wouldn't be to zero, but to the maximum over the collected history.
3. And lastly... I'm not sure how sensitive the algorithm is to removing (unused) lights... if it's not too much sensitive, maybe it would be possible to slowly decay the `max_cluster_model_lights` value. This would help when there's an extreme spike of dynamic lights in a cluster that stays visible a long time (eg some outdoor area).

Maybe you have an opinion on them? Or perhaps you can see a flaw I can't?

PR #234, opened 2022-07-17
 Keep a small history of 'light counts' for each cluster used for sampling lights (for #226)

res2k:
This is a different approach at addressing #226 after the changes #227 fell short in other scenarios.

I introduced a small _history_ of "cluster light counts" used for light selection.
Now, if light sampling is performed with an "old" RNG seed, we can pull out the light count used at that time from that history we're keeping around (based on the frame number, which is conveniently stored in the RNG seed as well).

This seems to work quite well - it fixes both the problem from #226 but crucially doesn't seem to make things worse if the number of lights in a cluster changes, as previously visible in e.g. the single-player entrance of the `security` map.
(Actually, I'd like to argue that `security` looks _better_ now, as now the "flashing" at the start, due to jumps in the max light count, is completely absent with this change!)

Other changes:
* Since we don't need the "max light count" any more, cluster lights are tightly packed (again)
* The frame number was _almost_ already contained in the RNG seed; I changed it to contain the unadulterated frame number, and for that I packed it a bit "tighter" as well.

### DIFF
--- a/src/refresh/vkpt/main.c
+++ b/src/refresh/vkpt/main.c
@@ -334,7 +334,7 @@ vkpt_destroy_all(VkptInitFlags_t destroy_flags)
 	if ((VKPT_INIT_DEFAULT & destroy_flags) == destroy_flags)
 	{
 		destroy_transparency();
-		vkpt_light_stats_destroy();
+		vkpt_light_buffers_destroy();
 	}
 
 	return VK_SUCCESS;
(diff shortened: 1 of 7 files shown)


res2k:
@apanteleev Ping?

apanteleev:
Merged - thank you.

And apologies about the long delay... I was mostly focusing on Portal with RTX in the last few months.

PR #235, opened 2022-07-17
Also store material original dimensions if no material definition is available

res2k:
Fixes incorrect scaling of textures w/o a custom material definition (ie custom textures on custom maps, most likely) after #222.


QuakeMM:
Just letting you know this fixed my bugs I was having with custom textures and old mods 😊❤️
quake006
quake007

PR #248, opened 2022-12-07
Fix up vulkan initialization

kkartaltepe:
Fix graphics/transfer queue selection to correctly pick the same queue on platforms that dont support split queues. this also preseve's the preference for split queues.

Remove unused request for sparseBindings, allows running on platforms that dont support it like mesa's ANV driver.

Without these fixes initialization fails either due to missing a transfer queue or missing sparseBinding on mesa's ANV driver for Arc GPUs.


apanteleev:
Merged with an additional bug fix. Thanks!

PR #250, opened 2022-12-13
Add XDG Support.

### LINKED ISSUE #243: Does not follow XDG specification on Linux.
Not very fun seeing more . folders appear in my home drive, at least set it to use XDG config or XDG /local/share

Connor-GH:
I was cleaning my HOME (~/) today and noticed a ``.quake2rtx`` residing there. I dove into Q2RTX's source code and round out that this path is hardcoded. This patch changes this behavior to rely on XDG_DATA_HOME and move the previous data there as well (if there is any).

Fixes #243.


res2k:
The home directory logic is probably inherited from Q2PRO, perhaps also check how it's used there.

src/unix/system.c
```diff
@@ -341,7 +344,16 @@ void Sys_Init(void)
     if (!homedir) {
 	    Sys_Error("Homedir not found!\n");
     }
-    sprintf(homegamedir, "%s/%s", homedir, ".quake2rtx");
+    xdg_data_home_dir = getenv("XDG_DATA_HOME");
+    if (xdg_data_home_dir == NULL) {
+	    xdg_data_home_dir = malloc(sizeof(char) * 200);
```
res2k, line 349: You know the length you need to assemble a homedir, no need to arbitrarily choose an allocation size.

src/unix/system.c
```diff
@@ -341,7 +344,16 @@ void Sys_Init(void)
...
+	    malloc_needed = 0;
+	    sprintf(xdg_data_home_dir, "%s/%s", homedir, ".local/share");
```
res2k, line 351: Always use snprintf

src/unix/system.c
```diff
@@ -341,7 +344,16 @@ void Sys_Init(void)
...
+    }
+    check_homegamedir = sprintf(homegamedir, "%s/%s", xdg_data_home_dir, "quake2rtx");
```
res2k, line 353: snprintf

src/unix/system.c
```diff
@@ -341,7 +344,16 @@ void Sys_Init(void)
...
+    if (malloc_needed && (check_homegamedir > 0)) {
+	    free(xdg_data_home_dir);
```
res2k, line 355: Freeing xdg_data_home_dir shouldn't be bound to whether the previous sprintf() succeeded or not.

setup/q2rtx.sh
```diff
@@ -2,21 +2,28 @@
...
+
 # If the game is installed via a package manager q2rtx won't be in the same
 # directory as q2rtx.sh
 if [[ -d "/usr/share/quake2rtx" ]]; then
 	BIN_PREFIX="/usr/share/quake2rtx/bin"
 fi
 
+if [[ -d "${HOME}/.quake2rtx" ]]; then
+        mv "${HOME}/.quake2rtx" "${XDG_DATA_HOME}/quake2rtx"
+fi
```
aaronp24, line 15: This opens a can of worms for anyone switching between old and new versions of q2rtx. Running an old copy after this mv will recreate ~/.quake2rtx, and then running a new version with this change will end up moving the old directory into a subdirectory of the new directory.

At the very least, this change should check whether `${XDG_DATA_HOME}/quake2rtx` exists before moving the old directory. But this will still create a mess for anyone who ever downgrades across this change.

setup/q2rtx.sh
```diff
@@ -2,21 +2,28 @@
 
 BIN_PREFIX="."
 
+XDG_DATA_HOME := ${HOME}/.local/share
```
aaronp24, line 5: This will override any user-supplied value for `XDG_DATA_HOME`. The specification says,
```
$XDG_DATA_HOME defines the base directory relative to which user-specific data files
should be stored. If $XDG_DATA_HOME is either not set or empty, a default equal to 
$HOME/.local/share should be used. 
```

src/unix/system.c
```diff
@@ -307,8 +308,11 @@ Sys_Init
 */
 void Sys_Init(void)
 {
+    _Bool   malloc_flag;
```
res2k, line 311: Shouldn't this be named `malloc_needed`?
Also, it should be initialized to 0/`false`.

src/unix/system.c
```diff
@@ -341,7 +345,25 @@ void Sys_Init(void)
     if (!homedir) {
 	    Sys_Error("Homedir not found!\n");
     }
-    sprintf(homegamedir, "%s/%s", homedir, ".quake2rtx");
+    xdg_data_home_dir = getenv("XDG_DATA_HOME");
+    if (!xdg_data_home_dir) {
+	    xdg_data_home_dir = malloc(sizeof(char) * PATH_MAX);
```
res2k, line 350: You could also use some equivalent to `strlen(homedir) + 1 + strlen(".local/share")` instead of `PATH_MAX`.
As I said previously, you actually already know the exact size you need.

src/unix/system.c
```diff
@@ -341,7 +345,25 @@ void Sys_Init(void)
...
+		    Sys_Error("xdg_data_home_dir: snprintf failed.\n");
+	    }
+    }
+    check_homegamedir = snprintf(homegamedir, sizeof(homegamedir), 
+				 "%s/%s", xdg_data_home_dir, "quake2rtx");
+    if (malloc_needed) {
+	    free(xdg_data_home_dir);
+    }
+
+    if (check_snprintf < 0) {
```
res2k, line 364: Wrong variable is checked, should be `check_homegamedir`.

res2k:
Did you actually try to compile your changes?
Because some of the things I commented on would actually trigger a compile error...

Connor-GH:
> Did you actually try to compile your changes? Because some of the things I commented on would actually trigger a compile error...

I did compile my changes, but only on a local test file. I figured it was an easy fix so I didn't git clone and just edited directly into the github editor.

Connor-GH:
@aaronp24 mentioned a good point: using an older version would cause issues in regards to saves. I believe that this should be addressed next.

src/refresh/vkpt/shader/global_ubo.h
```diff
@@ -124,8 +121,11 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 	UBO_CVAR_DO(tm_hdr_peak_nits, 800.0) /* Exposure value 0 is mapped to this display brightness (post tonemapping) */ \
 	UBO_CVAR_DO(tm_hdr_saturation_scale, 100) /* HDR mode saturation adjustment, percentage [0..200], with 0% -> desaturated, 100% -> normal, 200% -> oversaturated */ \
 	UBO_CVAR_DO(ui_hdr_nits, 300) /* HDR mode UI (stretch pic) brightness in nits */ \
+	UBO_CVAR_DO(pt_particle_brightness, 100) /* particle brightness */ \
```
appetrosyan, line 124: Why change this?

src/refresh/vkpt/shader/global_ubo.h
```diff
@@ -75,22 +75,19 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 	UBO_CVAR_DO(pt_cameras, 1) /* switch for security cameras, 0 or 1 */ \
 	UBO_CVAR_DO(pt_direct_polygon_lights, 1) /* switch for direct lighting from local polygon lights, 0 or 1 */ \
 	UBO_CVAR_DO(pt_direct_roughness_threshold, 0.18) /* roughness value where the path tracer switches direct light specular sampling from NDF based to light based, [0..1] */ \
-	UBO_CVAR_DO(pt_direct_dyn_lights, 1) /* switch for direct lighting from local sphere lights, 0 or 1 */ \
```
runlevel5, line 78: I do not think this change is related to XDG support. Please revert the changes

runlevel5:
@Connor-GH wondering if you are still working on this Pull Request. Love to get them revised for another round of review

Connor-GH:
> @Connor-GH wondering if you are still working on this Pull Request. Love to get them revised for another round of review

I put it on hiatus for the longest time and I apologize it. I will prioritize this and try to get it merged.

setup/q2rtx.sh
Connor-GH: Fixed in 41ee020bdafcd931e61435d6314ff2ac27fc8202.

src/refresh/vkpt/shader/global_ubo.h
Connor-GH: This was for an unrelated issue regarding RADV. The change was reverted in 1a888343e5c9077c004a7297855b9c1c2f9a2687.

src/refresh/vkpt/shader/global_ubo.h
Connor-GH: Done: 1a888343e5c9077c004a7297855b9c1c2f9a2687.

Connor-GH:
I believe it is good for another review if @runlevel5 or whomever wants to look at it.

src/refresh/vkpt/shader/global_ubo.h
appetrosyan: These kinds of changes should **usually** be included in the PR description. I like to have one commit per change and have a list of those commits in the PR description.

setup/q2rtx.sh
```diff
@@ -2,21 +2,30 @@
 
 BIN_PREFIX="."
 
+if [[ -z "${XDG_DATA_HOME}" ]]; then
+	XDG_DATA_HOME="${HOME}/.local/share"
+fi
```
appetrosyan, line 7: Consider

```suggestion
XDG_DATA_HOME="${XDG_DATA_HOME:=${HOME}/.local.share}"
```

Personally I'd prefer also to  check if the directory actually exists as well, so also consider:
```suggestion
mkdir -p "${XDG_DATA_HOME:="${HOME}/.local/share"}"
```

appetrosyan: See stack overflow for explanation

https://stackoverflow.com/questions/2013547/assigning-default-values-to-shell-variables-with-a-single-command-in-bash

src/unix/system.c
```diff
@@ -341,7 +345,27 @@ void Sys_Init(void)
     if (!homedir) {
 	    Sys_Error("Homedir not found!\n");
     }
-    sprintf(homegamedir, "%s/%s", homedir, ".quake2rtx");
+    xdg_data_home_dir = getenv("XDG_DATA_HOME");
+    if (!xdg_data_home_dir) {
+	    xdg_data_home_dir = malloc(sizeof(char) * 
+				       (strlen(homedir) + strlen("/.local/share") + 1));
+	    malloc_needed = 1;
```
appetrosyan, line 352: You're not actually checking if the `malloc` succeeded, and can potentially double-free. Set this in a ternary

src/unix/system.c
```diff
@@ -341,7 +345,27 @@ void Sys_Init(void)
     if (!homedir) {
 	    Sys_Error("Homedir not found!\n");
     }
-    sprintf(homegamedir, "%s/%s", homedir, ".quake2rtx");
+    xdg_data_home_dir = getenv("XDG_DATA_HOME");
+    if (!xdg_data_home_dir) {
+	    xdg_data_home_dir = malloc(sizeof(char) * 
+				       (strlen(homedir) + strlen("/.local/share") + 1));
```
appetrosyan, line 351:
```suggestion
				       (strlen(homedir) + strlen("/.local/share") + strlen("\0")));
```
More explicit and no need for a comment.

src/unix/system.c
```diff
@@ -341,7 +345,27 @@ void Sys_Init(void)
     if (!homedir) {
 	    Sys_Error("Homedir not found!\n");
     }
-    sprintf(homegamedir, "%s/%s", homedir, ".quake2rtx");
+    xdg_data_home_dir = getenv("XDG_DATA_HOME");
+    if (!xdg_data_home_dir) {
```
appetrosyan, line 349:
```suggestion
    if (!xdg_data_home_dir) {
        size_t len = strlen(homedir) + strlen("/.local/share") + 1;
```

You use this more than twice, and any desync can cause a memory leak.

src/unix/system.c
```diff
@@ -341,7 +345,27 @@ void Sys_Init(void)
...
+	    xdg_data_home_dir = malloc(sizeof(char) * 
+				       (strlen(homedir) + strlen("/.local/share") + 1));
+	    malloc_needed = 1;
+	    check_snprintf = snprintf(xdg_data_home_dir, 
+				      strlen(homedir) + strlen("/.local/share") + 1, 
+		     "%s/%s", homedir, ".local/share");
+            if (check_snprintf < 0) {
+		    Sys_Error("xdg_data_home_dir: snprintf failed.\n");
+	    }
```
appetrosyan, line 358:
```suggestion
        if (check_snprintf < 0) 
		    Sys_Error("xdg_data_home_dir:`XDG_DATA_HOME` not set, and and storing the path failed.\n");
```

Although removing the curly braces is a matter of taste, the error message is non-negotiable. It needs to explain what went wrong to a user.

src/unix/system.c
appetrosyan: Also, it should be 
```suggestion
	    mem_was_allocated = 1;
```
or 
```suggestion
	    need_to_free = 1;
```

src/unix/system.c
```diff
@@ -341,7 +345,27 @@ void Sys_Init(void)
...
+    }
+    check_snprintf = snprintf(homegamedir, sizeof(homegamedir), 
+				 "%s/%s", xdg_data_home_dir, "quake2rtx");
+    if (malloc_needed) {
+	    free(xdg_data_home_dir);
+    }
+
+    if (check_snprintf < 0) {
+            Sys_Error("homegamedir: snprintf failed.\n");
```
appetrosyan, line 367: Same comment as above about error message.

src/unix/system.c
res2k: > More explicit and no need for a comment.

This doesn't work as you think. `strlen("\0")` will return 0 (after all, there's nothing but a null terminator in that string).
So better just keep the constant 1.

appetrosyan: My bad!

src/unix/system.c
Connor-GH: This is a matter of taste, but not a necessary one I believe. The file is filled with mixed styles relating to this. At first, I thought it was curly braces if in a block (not including functions), but then there are examples of ``if`` statements with braces and without nesting. I don't think this is necessary.

Connor-GH: Improved error message in 7ae9ce53.

src/unix/system.c
Connor-GH: Fixed indentation in 7ae9ce53.

src/unix/system.c
Connor-GH: Done: 7ae9ce53.

src/unix/system.c
Connor-GH: 7ae9ce53

setup/q2rtx.sh
Connor-GH: It can be debated whether this is more concise or faster but either way, the directory needs to be made. 7ae9ce53.

appetrosyan (approved):

runlevel5 (approved):

Connor-GH:
I think it would be best for me to rebase and sync with NVIDIA/Q2RTX:master before this is accepted. I doubt anyone wants these many commits in their log or in their journey when they go to do ``git bisect`` in the future.

appetrosyan:
@Connor-GH 

> I'd also recommend squashing your commits into one. These changes are tiny compared to what the other commits do.

runlevel5:
> @Connor-GH I'd also recommend squashing your commits into one. These changes are tiny compared to what the other commits do

Or use the Squash and Merge function of Github

apanteleev:
Hi - I've asked someone more familiar with Linux than I am to take a look at this PR. 
I'll handle the merge and squash when he approves.

Connor-GH:
> > @Connor-GH I'd also recommend squashing your commits into one. These changes are tiny compared to what the other commits do
> 
> Or use the Squash and Merge function of Github
Of git* (``git merge --squash ...``) :)

aaronp24:
> Hi - I've asked someone more familiar with Linux than I am to take a look at this PR. I'll handle the merge and squash when he approves.

I'm generally not a huge fan of the Squash and Merge button because reviewing the commit message is just as important as the rest of the change, and if I understand correctly, doing that here would create a single commit with the description, "Add XDG Support".

apanteleev:
I'm not a fan of using GitHub UI for merging anyway, and I handle all merges locally, then push. If y'all want, I can set the squashed commit message to any descriptive text.

Connor-GH:
I removed an unneeded boolean in my code in b5326177.

Connor-GH:
> I'm not a fan of using GitHub UI for merging anyway, and I handle all merges locally, then push. If y'all want, I can set the squashed commit message to any descriptive text.

Perhaps the descriptionof the commit should tell users what the "xdg" spec is. The title should also specify that it is conforming to a specification, and that this should not be seen as a user-facing feature. https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html names "XDG" as the "XDG Base Directory Specification". The description should clarify that the game's data is to be stored in $XDG_DATA_HOME, since this path seems the best fit to put game data in. There should be a mention to the above freedesktop webpage if possible as well.

Lastly, this commit should mention that old game data will be migrated over upon next program execution.

src/unix/system.c
```diff
@@ -341,7 +344,32 @@ void Sys_Init(void)
...
+					"failed.\n");
+	    }
+    }
+    check_snprintf = snprintf(homegamedir, sizeof(homegamedir),
+				 "%s/%s", xdg_data_home_dir, "quake2rtx");
+
+	free(xdg_data_home_dir);
+
+    if (check_snprintf < 0) {
+           Sys_Error("homegamedir: snprintf failed.\n");
```
liam-middlebrook, line 371: Please log the non-zero `check_snprintf` value

liam-middlebrook (commented):
Indentation seems inconsistent between tab/space usage. Not sure how much of that is due to the GitHub UI though.

Functionally this looks fine to me, one small change requested. I don't plan to re-review before @apanteleev merges.

src/unix/system.c
Connor-GH: Ah, earlier code seems to use all spaces. I'll quickly fix this.

Connor-GH: > I am not sure if you were aware of the other whitespace errors in the file, but I went ahead and fixed them too.

This is a mistake on my end. Code style reflects 8 spaces in each conditional.

Connor-GH:
My commits have gotten wildly out of hand but they should be fine once squashed. I apologize for the mess that has been my commit history.

apanteleev:
Merged as 8cb6c6701b7d2818cf5610ade17618a53e43baf8

Thanks all, closing the PR.

Issue #231, opened 2022-06-04
Linux - Can't link after compilation - SDL 2 and XWayland issue?


BielBdeLuna:
```
[ 95%] Linking CXX executable ../../q2rtx
/usr/bin/ld: ../extern/SDL2/libSDL2.a(SDL_waylandvideo.c.o): in function `wl_display_get_registry':
/home/biel/code/quake/quake2/Q2RTX/build/extern/SDL2/wayland-generated-protocols/wayland-client-protocol.h:1032: undefined reference to `wl_proxy_marshal_flags'
/usr/bin/ld: ../extern/SDL2/libSDL2.a(SDL_waylandvideo.c.o): in function `xdg_wm_base_destroy':
/home/biel/code/quake/quake2/Q2RTX/build/extern/SDL2/wayland-generated-protocols/xdg-shell-client-protocol.h:482: undefined reference to `wl_proxy_marshal_flags'
/usr/bin/ld: /home/biel/code/quake/quake2/Q2RTX/build/extern/SDL2/wayland-generated-protocols/xdg-shell-client-protocol.h:482: undefined reference to `wl_proxy_marshal_flags'
/usr/bin/ld: ../extern/SDL2/libSDL2.a(SDL_waylandvideo.c.o): in function `wl_registry_bind':
/home/biel/code/quake/quake2/Q2RTX/build/extern/SDL2/wayland-generated-protocols/wayland-client-protocol.h:1145: undefined reference to `wl_proxy_marshal_flags'
/usr/bin/ld: /home/biel/code/quake/quake2/Q2RTX/build/extern/SDL2/wayland-generated-protocols/wayland-client-protocol.h:1145: undefined reference to `wl_proxy_marshal_flags'
/usr/bin/ld: ../extern/SDL2/libSDL2.a(SDL_waylandvideo.c.o):/home/biel/code/quake/quake2/Q2RTX/build/extern/SDL2/wayland-generated-protocols/wayland-client-protocol.h:1145: more undefined references to `wl_proxy_marshal_flags' follow
collect2: error: ld returned 1 exit status
make[2]: *** [src/CMakeFiles/client.dir/build.make:2952: ../q2rtx] Error 1
make[1]: *** [CMakeFiles/Makefile2:978: src/CMakeFiles/client.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
```

res2k:
This is an SDL issue: https://github.com/libsdl-org/SDL/issues/5088
[...]

PR #252, opened 2022-12-25
Update SDL2 to 2.26.1

res2k:
Hopefully addresses #231.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,6 +99,10 @@ ELSEIF(CMAKE_SYSTEM_PROCESSOR STREQUAL "AMD64" OR CMAKE_SYSTEM_PROCESSOR STREQUA
 
 ENDIF()
 
+# Work around CMAKE_INSTALL_PREFIX being a relative directory when using -B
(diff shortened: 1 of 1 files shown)


apanteleev:
GitHub [shows](https://github.com/libsdl-org/SDL/compare/fd196c446a476282dd457fbafe9e9b3feae3f370...f17058b562c8a1090c0c996b42982721ace90903) that the only differences between the two versions of SDL2 is the version number - is this right? How can that fix anything?

res2k:
If you manually follow the submodule directory you get here: https://github.com/libsdl-org/SDL/tree/fd196c446a476282dd457fbafe9e9b3feae3f370
It's latest commit is from Apr 2020, and by inspecting e.g. `configure.ac` you can see this is SDL 2.0.13.

Not sure how GH computed that diff.
Above what's shows there's the message "fd196c446a476282dd457fbafe9e9b3feae3f370 and f17058b562c8a1090c0c996b42982721ace90903 are entirely different commit histories. " - I don't think it can provide anything sensible and maybe falls back to commit diff only.

apanteleev:
I see, thanks. I should've looked at the diffs locally / more carefully.

Merged now.

PR #259, opened 2023-01-12
Add support for UNORM RGB8 output surface formats 

res2k:
Accepting them during surface picking is not enough, we also need to do a `linear_to_srgb()` on the color value written for display.

Should help with https://github.com/NVIDIA/Q2RTX/issues/217, as Wayland on NVIDIA currently only exposes `UNORM` output surfaces.

### DIFF
--- a/src/refresh/vkpt/draw.c
+++ b/src/refresh/vkpt/draw.c
@@ -28,8 +28,9 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 #include "shader/global_textures.h"
 
 enum {
-	STRETCH_PIC_SDR,
-	STRETCH_PIC_HDR,
+	STRETCH_PIC_SDR_WRITE_LINEAR,
+	STRETCH_PIC_SDR_WRITE_SRGB,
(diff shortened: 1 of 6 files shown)


apanteleev:
Thanks for the PR. 

Have you tried just creating the swap chain image views with an _SRGB format (while the swap chain itself is just _UNORM) instead of doing conversions in the shaders? I think that should work, at least on DX.

If not, I'll just merge this as-is, the code looks good.

res2k:
> Have you tried just creating the swap chain image views with an _SRGB format (while the swap chain itself is just _UNORM) instead of doing conversions in the shaders? I think that should work, at least on DX.

Hm, no, that didn't occur to me. I'll give it a try.

res2k:
> > Have you tried just creating the swap chain image views with an _SRGB format (while the swap chain itself is just _UNORM) instead of doing conversions in the shaders? I think that should work, at least on DX.
> 
> Hm, no, that didn't occur to me. I'll give it a try.

It works nicely. I created a new PR (#261) with that change.

PR #261, opened 2023-01-27
Add support for UNORM RGB8 output surface formats

res2k:
We can accept them during surface picking, but we need to set an SRGB format on the swapchain image view in that case.

Should help with https://github.com/NVIDIA/Q2RTX/issues/217, as Wayland on NVIDIA currently only exposes UNORM output surfaces.

### DIFF
--- a/src/refresh/vkpt/main.c
+++ b/src/refresh/vkpt/main.c
@@ -562,6 +562,8 @@ static bool pick_surface_format_sdr(picked_surface_format_t* picked_fmt, const V
 	} acceptable_formats[] = {
 		{VK_FORMAT_R8G8B8A8_SRGB, VK_FORMAT_R8G8B8A8_SRGB},
 		{VK_FORMAT_B8G8R8A8_SRGB, VK_FORMAT_B8G8R8A8_SRGB},
+		{VK_FORMAT_R8G8B8A8_UNORM, VK_FORMAT_R8G8B8A8_SRGB},
+		{VK_FORMAT_B8G8R8A8_UNORM, VK_FORMAT_B8G8R8A8_SRGB},
 	};
 
 	for(int i = 0; i < LENGTH(acceptable_formats); i++) {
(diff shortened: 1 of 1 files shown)


src/refresh/vkpt/vkpt.h
```diff
@@ -182,6 +182,7 @@ typedef struct QVK_s {
 	VkSwapchainKHR              swap_chain;
 	VkSurfaceFormatKHR          surf_format;
 	bool                        surf_is_hdr;
+	bool                        surf_write_srgb;
```
apanteleev, line 185: Looks like this is unnecessary and just left over from the first PR.

src/refresh/vkpt/main.c
```diff
@@ -537,24 +545,30 @@ static bool pick_surface_format_hdr(VkSurfaceFormatKHR* format, const VkSurfaceF
...
+static bool pick_surface_format_sdr(picked_surface_format_t* picked_fmt, const VkSurfaceFormatKHR avail_surface_formats[], size_t num_avail_surface_formats)
 {
-	VkFormat acceptable_formats[] = {
-		VK_FORMAT_R8G8B8A8_SRGB, VK_FORMAT_B8G8R8A8_SRGB,
+	struct {
+		VkFormat format;
+		VkFormat swapchain_view_fmt;
+	} acceptable_formats[] = {
+		{VK_FORMAT_R8G8B8A8_SRGB, VK_FORMAT_R8G8B8A8_SRGB}, {VK_FORMAT_B8G8R8A8_SRGB, VK_FORMAT_B8G8R8A8_SRGB},
+		{VK_FORMAT_R8G8B8A8_UNORM, VK_FORMAT_R8G8B8A8_SRGB}, {VK_FORMAT_B8G8R8A8_UNORM, VK_FORMAT_B8G8R8A8_SRGB},
```
apanteleev, line 564: Suggest splitting these lines so that there is one entry per line.

src/refresh/vkpt/vkpt.h
res2k: Duh, yes

src/refresh/vkpt/main.c
res2k: Sure

PR #262, opened 2023-01-28
Fix primitive overflow with pt_bsp_sky_lights > 1

res2k:
`NODRAW` surfaces would be double-collected by collecting with `filter_nodraw_sky_lights()`, exceeding the amount of allocated primitives.


res2k:
Hmm, no, this is not sufficient. If I mix sky with and without NODRAW set I still get a primitive overflow.

PR #269, opened 2023-02-11
GH workflow: Add prerequisites to allow Wayland support

res2k:
This change was originally part of #221 earlier (discussed in https://github.com/NVIDIA/Q2RTX/pull/221#issuecomment-1124440829), but was somehow lost along the way.
Anyway, (re-)adding those dependencies makes the workflow builds support Wayland.


res2k:
Closing, as with 5592ef1 the necessary libs to provide Wayland support are available as well.

Issue #272, opened 2023-04-03
No music on Linux when placed in game directory


abalfoort:
Placed xx.ogg files in [gamedir]/baseq2/music/

Startup message:
`No Ogg Vorbis music tracks have been found, so there will be no music.`

In console:
`ogg info`
outputs:
```
Tracks:
Total: 1 Ogg/Vorbis files.
State: Stopped.
```

`dir music`
outputs:
```
02.ogg
03.ogg
04.ogg
05.ogg
06.ogg
07.ogg
08.ogg
09.ogg
10.ogg
11.ogg
12.ogg
13.ogg
14.ogg
15.ogg
16.ogg
17.ogg
18.ogg
19.ogg
20.ogg
21.ogg
20 files listed
```

If I create a symlink:
`ln -s [gamedir]/baseq2/music $HOME/.quake2rtx/baseq2/music`

Music is being played.

https://github.com/NVIDIA/Q2RTX/blob/master/src/client/sound/ogg.c:
I see that fs_gamedir is always pointing to  $HOME/.quake2rtx. 
It is overwritten in common/files.c (3450):
```
            add_game_dir(FS_PATH_BASE | FS_PATH_GAME,
                         "%s/"BASEGAME, sys_homedir->string);
```
What if we save the current fs_gamedir in a new variable. fs_gamedir_linux and use that to add to potMusicDirs?

runlevel5:
@abalfoort you could close this issue now

PR #273, opened 2023-04-04
Fix: No music on Linux when placed in game directory

abalfoort:
This is a fix for issue #272


Calinou (changes requested):
Please keep the existing indentation style to make the diff smaller.

abalfoort:
I reverted the indentation commit but it doesn't make it more readable. Also, this file already has a mix of tabs and spaces indentation while most, if not all, other files use spaces for indentation.

apanteleev:
Merged with squash to keep the history cleaner. https://github.com/NVIDIA/Q2RTX/commit/12cd2a38da02bb3c980fc7d5d114dc275e7b5c1d

PR #284, opened 2023-07-03
Sync with Q2PRO: Misc 'shared' stuff

res2k:
Various changes involving `shared.{c,h}`.


Paril:
@res2k Not a major issue, but, I've noticed that `RF_NOSHADOW` has different values between q2rtx and q2pro now; it might be worth doing another general sync (this change was from 2022 https://github.com/skullernet/q2pro/commit/ffc80035ede785d2987120ae1b236ff0cb88eec3 so there might be other things missing)

PR #288, opened 2023-07-17
Performance measurement bits

res2k:
Some tidbits to help with performance measuring:
* Display a moving average for profiler values
* Add ability to run a timedemo multiple times, back-to-back (useful to perform benchmark runs one after another)
* Add profiler measurements for individual indirect lighting bounces


apanteleev:
I noticed a couple issues with the new profiler feature:
- The "update environment" line is flickering between 0.00 and N/A. this is easy to fix by comparing the values to a small threshold like 0.005 instead of 0.
- Average values are not reset when the pass that produced them ceases to execute. Like, change `pt_num_bounce_rays` to 0 and see the indirect passes showing N/A as the immediate value but sticking to the old accumulated value.

res2k:
> I noticed a couple issues with the new profiler feature:
> 
>     * The "update environment" line is flickering between 0.00 and N/A. this is easy to fix by comparing the values to a small threshold like 0.005 instead of 0.
> 
>     * Average values are not reset when the pass that produced them ceases to execute. Like, change `pt_num_bounce_rays` to 0 and see the indirect passes showing N/A as the immediate value but sticking to the old accumulated value.

Addressed both.

PR #290, opened 2023-07-20
Apply distortion effect on TRANS walls

res2k:
...to make them look a bit more interesting.
Some maps use a TRANS flag with various "random" textures which are not intended to be used as "glass" and don't have the GLASS material kind set, and even with that, may look bad due to their colors not working well as "throughput" colors.
So keep them acting as a "non-physical see-through material", but apply some distortion to the "through" ray.

Before:
quake001
After:
quake000
(Location is in xatrix' xcompnd1 map.)


apanteleev:
Nice idea, but this change breaks `cl_gunalpha`.

res2k:
> Nice idea, but this change breaks `cl_gunalpha`.

Fixed.

PR #308, opened 2023-08-24
Add more entity shell types

res2k:
Adds support for renderfx shell types beyond combinations of red/green/blue.
Includes all shell types used by expansions or the rerelease version.


src/refresh/vkpt/shader/global_ubo.h
```diff
@@ -266,8 +267,7 @@ BEGIN_SHADER_STRUCT( ModelInstance )
 	int iqm_matrix_offset_curr_frame;
 	int iqm_matrix_offset_prev_frame;
 
-	int frame;
-	float alpha;
+	uint frame_and_alpha;
```
apanteleev, line 270: I was a bit surprised to find that 'frame' is in the high 16 bits and 'alpha' is in the low 16 bits. I usually name combined variables like that in the opposite order, so `alpha_and_frame`, see e.g. `emissive_and_alpha` in `bsp_mesh.c`. Please rename and (ideally) add a comment explaining the packing.

res2k: Somehow that pattern totally went past me. Addressed.

apanteleev: No worries, and thanks for fixing it!

PR #310, opened 2023-08-24
Rename the Linux artefact output

runlevel5:
To denote that the binaries are for Linux x86_64. Because there are many 64bit architectures for Linux right now, personally I find `x64` can be confusing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,7 +49,7 @@ jobs:
           license.txt
           notice.txt
 
-  build-linux-x64:
+  build-linux-x86_64:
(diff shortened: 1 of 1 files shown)


apanteleev:
I think this is unnecessary, `x64` is a common alias for `x86-64` and nothing else. See e.g. https://en.wikipedia.org/wiki/X86-64

runlevel5:
`x86_64` is more commonly used across Linux distro (ref: `uname -m`). IMHO by adhering to that convention, it would help ease the effort of package maintainer in writing installation scripts to pick up right artefacts.

PR #316, opened 2023-09-07
Fix an incorrect check 

res2k:
...introduced in 0f5ef84b9168ff483e9dd8a5ad43c19e74f1f83b

An input check condition required some tweaking which was missed when porting a commit.

### DIFF
--- a/src/common/msg.c
+++ b/src/common/msg.c
@@ -295,7 +295,7 @@ MSG_WriteBits
 */
 void MSG_WriteBits(int value, int bits)
 {
-    if (bits == 0 || bits < -31 || bits > 32) {
+    if (bits == 0 || bits < -31 || bits > 31) {
         Com_Error(ERR_FATAL, "MSG_WriteBits: bad bits: %d", bits);
     }
 


apanteleev:
Merged, but can you please explain what's the idea behind negative `bits`? I see its sign being removed right at the beginning of the function, but some places actually call `MSG_WriteBits` with negative `bits`. Why?

res2k:
I think the answer can be found in `MSG_ReadBits()`: There, a negative count of bits indicates "sign extension".
Say, you have the bits "0xf000". `MSG_ReadBits(16)` would return 0x0000f000, `MSG_ReadBits(-16)` would return 0xfffff000.
The same scheme is used for `MSG_WriteBits()`. While it doesn't change the results, using negative number communicates that a signed value is being written, so the corresponding `MSG_ReadBits()` for the data should also use a negative bit count, lest the value may not be interpreted correctly.

apanteleev:
That makes sense, thanks. It's quite an obscure solution though.

res2k:
It's simultaneously obscure yet somehow strangely sensible ("You want a signed value? Put a sign in front of it's bit size.")

PR #317, opened 2023-09-07
Sync with Q2PRO: Sound changes

res2k:
Lots of fixes and improvements.
A highlight is the new support for "underwater" sound filtering.

One significant omission is the Ogg support that Q2PRO gained in the meantime.
It's quite different from the current code taken from Yamagi... two significant differences are (1) current code uses stb_vorbis and (2) current code provides "track mapping" for the expansions.
I think both of those are worthwhile to keep, so I'll look into getting the Ogg code closer to the upstream Q2PRO version, but retain the mentioned features.


res2k:
@apanteleev , what do you think?

If the number of changes is too much, I can try to split it up into smaller chunks.
And/or if there are some commits you're concerned about I can try to separate those.

apanteleev:
Sorry about the delay, I saw this and then you added some more commits - maybe there were more changes coming?..

The changes look OK, I merged the PR, didn't read too deep into the code though. Sounds and music still work, underwater filter also works. Some things were added and then reverted, I mean the ADPCM support; some changes are irrelevant (the `LIBAL` change on Win32). But that's OK.

Issue #318, opened 2023-09-11
Weapon renders into BSP surfaces when offset clipping isn't enough.


WatIsDeze:
For example, in the current crouched position at the start of base1/demo1. Simply just having the weapon being 'offset' by game code, won't always cut it. Minor issue, and yes, occurs only if you try really hard. Possibly easy to fix?

(The game code for offsetting the weapon doesn't take the weapon animation into consideration, hence, this happens.)
![image](https://github.com/NVIDIA/Q2RTX/assets/10252874/13d4f0f3-7f16-4e28-ad7f-6dacff3da57d)

res2k:
The logic to adjust the gun position is in `CL_AdjustGunPosition()`. It's using some hardcoded "gun dimensions", though, which are probably not quite suitable in this case (and possibly others as well).

res2k:
It may look superficially simple to fix - just adjust the hardcoded "gun dimensions", right? - however, this may look certain _other_ weapons look bad (being wall offset without appearing close to a wall).
Not sure what the best approach would be. Maybe using the view model bbox could give satisfactory results.

WatIsDeze:
It's been long ago since I did OpenGL API stuff but I recall in those days there was a solution related to depth testing, which would prevent minor issues such as this one.

Actually peeking around a bit, Q2RTX its GL renderer responds to the RF_DEPTHHACK flag. However there is zero results of finding any records of this flag being checked for in the VKPT code.


Found in shared.h
`#define RF_DEPTHHACK        16      // for view weapon Z crunching`

Found in client/entities.c
`gun.flags = RF_MINLIGHT | RF_DEPTHHACK | RF_WEAPONMODEL;`

Responded to/in/by `GL_DrawAliasModel` using 'glDepthRange'.

res2k:
The RTX renderer works differently, it doesn't have a depth hack equivalent.
But even if it would have, it probably would result in looking weird in cases where the hack would kick in:
The view weapon is actually properly located in the world, so reflections, lighting etc look reasonably right.
Now, cases where the depthhack would be in effect, the view weapon would be inside a wall or such - resulting in unexpected lighting and reflections, probably looking quite off. So moving the whole gun makes sense, at least for the RTX renderer.

apanteleev:
[...]

PR #319, opened 2023-09-13
Scale the view weapon to prevent intersection with walls

apanteleev:
Following up on https://github.com/NVIDIA/Q2RTX/issues/318

### DIFF
--- a/src/client/client.h
+++ b/src/client/client.h
@@ -518,6 +518,7 @@ extern char        cl_cmdbuf_text[MAX_STRING_CHARS];
 // cvars
 //
 extern cvar_t    *cl_gunalpha;
+extern cvar_t    *cl_gunscale;
 extern cvar_t    *cl_predict;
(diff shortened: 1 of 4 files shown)


res2k:
Shouldn't the `v_flareg` case be adjusted to `gun.scale *= 0.3f;` (`*=` instead of `=`)?

Generally seems to work well. I agree to the lighting differences not really being noticeable "unless you look for it".

If you set more extreme scale values (eg 0.01), the gun isn't just "slightly shifting on screen", it's doing quite the funny stuff.
Might be interesting to see why that is... but the current default generally looks fine, so IMO some explanation would just be "nice to have".

res2k:
@WatIsDeze, if you could also look how this behaves in the corner case(s) you found, that would be grand!

apanteleev:
Changes were made, thanks for the feedback @res2k

PR #322, opened 2023-09-21
Sync with Q2PRO: Ogg playback

res2k:
Q2RTX got it's Ogg playback code from Yamagi Quake 2, but since the last merge of Q2PRO changes, the latter also got Ogg playback support - but a different one, of course.

But Q2PRO also received some enhancements: Loading Ogg files for sound effects, and specifying music track names (not just numbers).
One notable difference: Q2PRO supports reading music in archives; stb_vorbis only supports stdio-based file access.
This could be addressed by reading the complete ogg file into memory, or using the stb_vorbis "pushdata" API, but neither has been implemented.

This PR attempts to bring the Ogg playback code closer to what's currently found in Q2PRO, while retaining some Yamagi-isms, mainly the track number mapping for the GOG tracks and using stb_vorbis instead of libogg/libvorbis for playback.


apanteleev:
Looks OK, but here's one odd thing I noticed while testing these changes:
```
> ogg info
Tracks:
 - 02 W:/q2rtx/baseq2/music/Track02.ogg
 - 03 W:/q2rtx/baseq2/music/Track03.ogg
 - 04 W:/q2rtx/baseq2/music/Track04.ogg
...
 - 19 W:/q2rtx/baseq2/music/Track19.ogg
 - 20 W:/q2rtx/baseq2/music/Track20.ogg
 - 21 W:/q2rtx/baseq2/music/Track21.ogg
Total: 20 Ogg/Vorbis files.
State: Playing file W:/q2rtx/baseq2/music/Track19.ogg at 940480 samples.

> ogg play 21
OGG_PlayTrack: 21 out of range.
```

Tracks 0 and 1 are invalid because of CD track rules, that's fine. But it looks like `trackcount` that's used to determine that the provided index is correct is only the length of a 0-based track list, which is obviously missing tracks 0 and 1.

res2k:
The track number check came from Yamagi, but it doesn't really make sense any more, as the tracklist is now tightly packed and `trackcount` isn't the "maximum track number" any more. Removed in [2e50b76](https://github.com/NVIDIA/Q2RTX/pull/322/commits/2e50b76ebd19669532521a0d4936a2642b6ac06d).

PR #326, opened 2023-10-01
Sync with Q2PRO: Client-side gun tweaks

res2k:
Q2PRO added some other client side gun options (separate FOV, custom position), port those over.

For OpenGL, the "gun FOV" uses a different projection matrix when rendering the viewweapon.
For RTX rendering, we don't have that separate rendering pass, so address this by manipulating the viewweapon's entity matrix in such a way that it appears as if rendered with a different FOV.
Also seems to work with cylindrical projection.


apanteleev:
The FOV thing seems to work well, but the new `cl_gun_{x,y,z}` variables don't work so well with `cl_gunscale`. When the gun scale is small, the apparent offset of the gun is larger than it should be. So, either gun scaling should be implemented as part of the hacky matrix, or the offset settings should be removed, I think.

res2k:
I'll take a look.

res2k:
Just multiplying the gun scale into the offset values seems to work nicely.

PR #327, opened 2023-10-01
Sync with Q2PRO: Memory allocation

res2k:
A bunch of changes to the memory allocation functions.

### DIFF
--- a/src/common/zone.c
+++ b/src/common/zone.c
@@ -84,8 +84,7 @@ static inline void Z_CountAlloc(const zhead_t *z)
 }
 
 #define Z_Validate(z) \
-    Q_assert(z->magic == Z_MAGIC); \
-    Q_assert(z->tag != TAG_FREE);
+    Q_assert((z)->magic == Z_MAGIC && (z)->tag != TAG_FREE)
 
 void Z_LeakTest(memtag_t tag)
 {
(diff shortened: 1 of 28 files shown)


src/common/zone.c
```diff
@@ -125,14 +128,29 @@ void Z_Free(void *ptr)
...
         free(z);
     }
 }
 
+/*
+========================
+Z_Freep
+========================
+*/
+void Z_Freep(void *ptr)
```
apanteleev, line 143: Suggest changing the parameter type to `void **p` to make it clearer what `Z_Freep` is doing. And fixing a couple of warnings about type mismatches where this function is used to deallocate arrays of pointers.

res2k: Done.
The clarity of what it does is a good point, but I suspect the additional warnings was the reason this is `void*` instead of `void**`.
They can be avoided using casts (which I added), but I think having to use one for every call to `Z_Freep()` isn't strictly better than just having take a `void*`. (Dunno, maybe something with macros that could be done?)

apanteleev: Thanks!
I was only getting 2 warnings when I tried it though, using MSVC 2019. Which compiler are you using? Just curious.

res2k: gcc warns extensively. Version 11, in my case.

PR #334, opened 2023-10-06
Sync with Q2PRO: Game constification, formatting, code style

res2k:
These are game changes from Q2PRO which shouldn't affect any behavior.

* Constifications
* Formatting fixes
* Subdirectory with sources renamed to "game"


src/game/g_local.h
```diff
@@ -490,11 +472,10 @@ extern  int snd_fry;
 #define MOD_TRIGGER_HURT    31
 #define MOD_HIT             32
 #define MOD_TARGET_BLASTER  33
-#define MOD_FRIENDLY_FIRE   0x8000000
```
apanteleev, line 493: `0x8000000 != BIT(31)` (one 0 too short)
which one is correct?

res2k: The value in the original game sources is "one 0 too short".
However, I don't think the actual value really matters: AFAICS, the "means of death" isn't actually something that's persisted, so `MOD_FRIENDLY_FIRE` can be any flag value, as long as it doesn't conflict with any other `MOD_xxx` value - which is the case for either value.
(I suspect that, when this change was made in Q2PRO, either `0x8000000` was misread to contain an extra zero, or it was decided that it's "nicer" to use the MSB as the flag bit, hence the change to `BIT(31)`.)

apanteleev:
I merged your PR right before you force-pushed it, to rebase, I assume. I don't see any new commits or anything. Ok to close?

res2k:
The weirdest thing is happening to me - GitHub reports a conflict where there is none in the actual sources.
(The force push was an attempt to address this...)
Apparently you didn't suffer the "phantom conflict", since you were able to merge the PR - so it's good to close.

PR #344, opened 2023-10-14
Added several more projections (pt_projection 2, 3, and 4)

RandomDude4:
I started adding more projection to Q2RTX in 2020. I even made a video about it: https://youtu.be/aAJl7b4i_ic
But I never tried to merge my code with the original NVIDIA code back then. 

So this is my attempt at finally merging the code so everyone can experience one of the great benefits of Ray Tracing, which is its ability to draw any projection natively. 
Personally I prefer playing in the stereographic projection, as rectilinear has huge distortions when playing on high FOV or on an Ultra-wide monitor (but unfortunately it's the only option in rasterized games). Cylindrical is also not really suitable for gaming.

The new projection code is written in standard trigonometric form (as can be found in any text book/Wikipedia), with latitude and longitude, etc. Making it easy for others to understand and play around with it, and I also never noticed any performance difference from the standard rectilinear projection.

I was not sure about the "previous" variable used for the original rectilinear and cylindrical projections, it didn't seem to make a difference with or without out it, so I left it out for the new projections.

Other projections that may be interesting to add in the future: Panini, Hammer, Hybrid-projections (blends/mixing of projections, inspired by Flex FOV and other projects like it)


Fubaxiusz:
It would be nice to add Pantomorphic perspective too, which mixes azimuthal projections. https://github.com/Fubaxiusz/fubax-shaders/blob/master/Shaders/PerfectPerspective.fx

Fubaxiusz:
Actually it would be nice if Pantomorphic projection was a standard feature of RTX, since it's already practically used by thousands of players in games via ReShade. It gives practical advantage as certain projections provide perception of certain things accurately, like:

- preserves area (equisolid) = accurate perception of distance to things.
- preserves radial angle (equidistant) = accurate perception of speed across image.
- preserves shape (stereographic) = accurate perception of mental 3D space, great for navigation.
- preserves lines (rectilinear) = standard projection

Pantomorphic makes it possible to combine many of them. Take racing example, driver would like to perceive accurately speed, distance to the corner and a shape of the turn. The solution is equisolid projection for X axis (distance to corner), equidistant for bottom vertical axis for the speed of the road below and stereographic for top Y axis to see the turn-shape ahead. It was blindly tested and gave advantage.

Let me know NVIDIA.

apanteleev:
Hi @RandomDude4,

Thank you for the contribution. The changes generally look good, but there's a few things that I think can be improved:
- Move the computation of various projection parameters based on fov and aspect ratio onto the CPU, similar to how `cylindrical_hfov` is computed. I'm talking about things like `log(tan(M_PI * 0.25 + (vfov / 2.0) * 0.5));` Not sure if it's better to allocate UBO parameters for all projections, or have a few shared scalars and call them like `projection_param1` etc... Multiple parameters would look cleaner.
- Add constants for all projection modes and use them instead of literals in `switch(pt_projection) { case 0: ...; case 1: ...; }`
- Either rename `angleScale` to `ANGLE_SCALE` because it's a preprocessor define, or - better - change it to `const float c_AngleScale = 0.5f;`

I did some quick performance testing and didn't see any noticeable changes on GPUs from all 3 major vendors, so apparently, it's OK to add a bunch of extra conditions everywhere projection math is used.

@Fubaxiusz - Feel free to implement Pantomorphic projection on top of these changes - ideally when they're final and merged, and submit a pull request.

apanteleev:
One more thing about this:

> I was not sure about the "previous" variable used for the original rectilinear and cylindrical projections, it didn't seem to make a difference with or without out it, so I left it out for the new projections.

The "previous" parameters are needed for the denoiser and TAAU passes to work correctly when the FOV changes between the previous and current frames. It's a minor thing because you don't change FOV often (unless it's a zoom-in for a sniper rifle, for example), but minor things like that ensure a smooth experience without visual artifacts.

RandomDude4:
Thnks @apanteleev for the suggestions, I will implement them and update the pull request in the coming week.

RandomDude4:
I implemented all the suggestions by @apanteleev. I hope I did not missinterpret anything, but in that case just let me know and I will fix it.

I have yet to investigate and understand exactly how the "previous" parameters works work, and find all the places in the code that is affected etc. TBD.

apanteleev:
Thanks @RandomDude4 !

I made a few more cosmetic changes to your code, implemented the previous frame thing, and fixed the player setup view (that should always use rectilinear projection). Merged now.

PR #346, opened 2023-10-15
Sync with Q2PRO: UI stuff

res2k:
A bunch of upstream changes/improvements to the UI.

### DIFF
--- a/inc/client/ui.h
+++ b/inc/client/ui.h
@@ -35,7 +35,7 @@ void        UI_Shutdown(void);
 void        UI_ModeChanged(void);
 void        UI_KeyEvent(int key, bool down);
 void        UI_CharEvent(int key);
-void        UI_Draw(int realtime);
+void        UI_Draw(unsigned realtime);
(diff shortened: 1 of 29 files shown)


apanteleev:
Hi @res2k - thanks for yet another package of q2pro changes.
After merging this, I noticed that the status bar rendering was somewhat broken, so it's fixed in 9e0554b4115227cea473c58d198d70bf536bac06. You might want to follow up with q2pro to see if the same bug exists there.

res2k:
> After merging this, I noticed that the status bar rendering was somewhat broken

Oops

> You might want to follow up with q2pro to see if the same bug exists there.

The code in Q2PRO looks different around the line you changed; it does `value = cl.frame.ps.stats[value];` instead of `...[index]`; the latter is a Q2RTX-specific change introduced in https://github.com/NVIDIA/Q2RTX/commit/2d5c6a7837842bd499e51b10aaaf31c80078ef37.

PR #364, opened 2023-11-23
Fix gcc warning that fov_scale may be used uninitialized

res2k:
...which is actually true for `PROJECTION_CYLINDRICAL`.


RandomDude4:
Sorry, that's my fault.
The same should be true for "float rad_per_pixel" which is defined just about fov_scale (as it is not used in most of the switch cases), but for some reason I do not get the same warning for it.

PR #366, opened 2023-11-30
Added Panini Projection to pt_projection

RandomDude4:
Added the Panini projection, specifically the "cylindrical stereographic" version with the parameter D = 1.0

D can be any value in order to seamlessly scale between projections:
D = 0 gives the rectilinear projection
D = +inf should give the "cylindrical
orthographic projection"

More info in this paper on why Panini is natural for the human eye:
http://dx.doi.org/10.2312/COMPAESTH/COMPAESTH10/009-016

Question:
Perhaps the Panini and Stereographic projections should be given a lower pt_projection values, as these are the most useful next to rectilinear? Cylindrical, mercator, etc. are just for demonstration, but very disorienting to use in an actual game.

### DIFF
--- a/baseq2/q2rtx.menu
+++ b/baseq2/q2rtx.menu
@@ -133,7 +133,7 @@ begin video
     pairs "full screen display" vid_display $vid_displaylist
     toggle "vertical sync" vid_vsync
     pairs "FPS counter" scr_fps "off" 0 "show FPS" 1 "show FPS and resolution scale" 2
-    range -f "%.0f deg" "field of view" fov 60 120 5
+    range -f "%.0f deg" "field of view" fov 60 160 5
     blank
 
     ifeq vid_rtx 1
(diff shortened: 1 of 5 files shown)


apanteleev:
Thanks, I'll give it a try.
One thing stands out: it's spelled Pannini (double N). :)

> Perhaps the Panini and Stereographic projections should be given a lower pt_projection values, as these are the most useful next to rectilinear?

Ok, sure.

RandomDude4:
> One thing stands out: it's spelled Pannini (double N). :)
When I saw your comment I couldn't believe that I had managed to misspelled it all this time!? 
But after checking online, both Unity and Unreal Engine documentation spell it "Panini" with one "n", (fyi: they have only implemented it as a post process, not natively like here, so in their case it results in a blurred image)
https://docs.unity3d.com/Packages/com.unity.render-pipelines.universal@7.1/manual/Post-Processing-Panini-Projection.html
https://docs.unrealengine.com/4.26/en-US/RenderingAndGraphics/PostProcessEffects/PaniniProjection/

But you are right that the name of the original painter Gian Paolo Pannini (1691-1765), which the projection is named after, does have the double "nn".
Let me know which version you prefer, and I will update the code and documentation accordingly.

Do you have a preferred order of the projections in pt_projection? 
Rectilinear and Pannini are probably the most commonly found in gaming, stereographic is also popular as far as I know. The others are rarely used when playing games, more for texture mapping etc.

This is a suggestion:
pt_projection 0 -> rectilinear
pt_projection 1 -> panini
pt_projection 2 -> stereographic
pt_projection 3 -> cylindrical
pt_projection 4 -> equirectangular
pt_projection 5 -> mercator

apanteleev:
Regarding the spelling - indeed, I see the single-N and double-N versions used interchangeably in various sources. I don't have a strong preference, but since both Unity and Unreal use the single-N version, let's do the same.

Your suggestion for the projection order looks good.

RandomDude4:
I renumbered all the cases according to the suggestion above. 
I also noticed a redundant "default:" case in main.c that I removed.

PR #376, opened 2024-02-08
Sync with Q2PRO: Water warp effect

res2k:
Upstream added a "water warp" effect to the GL renderer. This adds the same effect to the RTX renderer.

The principle is the same; the water warp is applied as the last step after rendering the view.
I opted to do this in the "final blit" step; the "final_blit_lanczos" shader was turned into a more generic "final_blit" shader, with lanczos filtering and warping controllable via specialization constants. (It's now also used for a "simple blit", ie both filtering and warping is turned off.)

### DIFF
--- a/doc/client.md
+++ b/doc/client.md
@@ -1113,7 +1113,7 @@ Default value is 1 (enabled).
 
 #### `gl_waterwarp`
 Enable screen warping effect when underwater. Only effective when using
-GLSL backend. Default value is 1 (enabled).
+GLSL backend. Default value is 0 (disabled).
 
 #### `gl_fontshadow`
 Specifies font shadow width, in pixels, ranging from 0 to 2. Default value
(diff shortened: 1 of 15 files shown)


apanteleev:
Thanks @res2k - I'm not sure why someone would want to make playing underwater even more difficult, but if q2pro has this effect, why not. :)

I tested the effect, and it looks good, but there's one issue: when DRS is enabled and drs_maxscale > 100, there is a wavy black border at the right and bottom edges of the screen.

res2k:
> Thanks @res2k - I'm not sure why someone would want to make playing underwater even more difficult, but if q2pro has this effect, why not. :)

Actually, this effect was a feature of original Quake II, back when it shipped - albeit in the software renderer only, as it wasn't feasible with the 3D accelerators back then.
So you could see this as actually restoring a feature ;)

> I tested the effect, and it looks good, but there's one issue: when DRS is enabled and drs_maxscale > 100, there is a wavy black border at the right and bottom edges of the screen.

Fixed it with some clamping.

Also added menu entries for warping.

Lastly, I noticed some "border weirdness" when FSR is enabled... but this seems to be more an FSR thing than warping, so I'll look into that separately.

PR #379, opened 2024-02-18
Sync with Q2PRO: Assertive things

res2k:
More replacements of statements patterned `if(!some_check) Com_Error(ERR_DROP, ...)` with `Q_assert(some_check)`.

### DIFF
--- a/src/client/effects.c
+++ b/src/client/effects.c
@@ -57,9 +57,7 @@ void CL_SetLightStyle(int index, const char *s)
 
     ls = &cl_lightstyles[index];
     ls->length = strlen(s);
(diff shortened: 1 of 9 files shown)


src/client/effects.c
```diff
@@ -57,9 +57,7 @@ void CL_SetLightStyle(int index, const char *s)
 
     ls = &cl_lightstyles[index];
     ls->length = strlen(s);
-    if (ls->length > MAX_QPATH) {
```
apanteleev, line 60: This is not an equivalent replacement, should be `<=`

apanteleev: I looked again and I take it back. The new version is correct.

PR #381, opened 2024-02-20
Support E2K (MCST Elbrus 2000)

r-a-sattarov:
- Add default C flags for e2k
- Add library and runtime output name for e2k
- Update CPUSTRING for e2k

Ref: https://lvee.org/en/abstracts/303


apanteleev:
Thanks, this is interesting. I assume these changes make q2rtx compile for E2K, but does it actually run, with ray tracing? I haven't been following modern developments around E2K, and having drivers for modern GPUs seems to be a likely problem.

DJs3000:
> Thanks, this is interesting. I assume these changes make q2rtx compile for E2K, but does it actually run, with ray tracing? I haven't been following modern developments around E2K, and having drivers for modern GPUs seems to be a likely problem.

A 7900XT graphics card is now available on e2k. Mesa 23.3.5.

apanteleev:
Merged.

> My colleague Dmitriy Bachilo @dbachilo tried to run on Elbrus the game Half Life, which uses ray tracing.

Oh right, I think I saw that video when it came out. :)

dbachilo:
Right, that happened. Ivan Avdeev was kind to port his patches for xash3d on e2k and we successfully ran vulkan raytraced Half-Life on Elbrus 8CB with Radeon 6700. Lets make Q2RTX happen!

DJs3000:
Works. We still need to optimize.
I'll send you a link to the video later.

image

Issue #306, opened 2023-08-17
[Feature Request] Support compilation against shared system libs


runlevel5:
I could tell that Q2RTX has a list of bundled external dependencies residing in `./extern`. Most of them are packaged and support by Ubuntu and Fedora. I would like to propose to add `USE_SYSTEM_<LIB_NAME>` flags in CMakeList to toggle optional compilation against the system libs.
```[tasklist]
### Tasks
- [x] cURL
- [x] stb
- [x] openAL
- [x] zLib
- [x] SDL2
- [ ] Vulkan-Headers
- [ ] glslang
- [ ] tinyobjloader-c
```

Connor-GH:
I was actually just about to open a similar issue. Currently, on gentoo (with normal glibc), there exists a bug where compilation fails because of a zlib ``OF`` macro. This prevents the zlib library from building, and as a result also prevents Q2RTX from building. If the system library had been used, the gentoo maintainers' fix would have already been used to let the library build, and Q2RTX could just use that. 

(by the way, the fix for gentoo systems compiling zlib is to specify ``-DOF\(x\)=x`` at the command line.)

runlevel5:
If the core team is okay with this feature request, I could provide patches

apanteleev:
The project is discontinued; no new features or fixes will be implemented here. Closing.

PR #390, opened 2024-03-27
Add option to compile against shared OpenAL lib

runlevel5:
## Issue

https://github.com/NVIDIA/Q2RTX/issues/306

## Motivation

Having the flexibility to opt in shared system lib is a big plus for packaging

## Changes

* Add `USE_SYSTEM_OPENAL` option


src/CMakeLists.txt
```diff
@@ -471,6 +471,11 @@ target_compile_options(server PRIVATE "${WARN_MISSING_PROTOTYPES}")
 
 IF (TARGET client)
     TARGET_COMPILE_DEFINITIONS(client PRIVATE USE_CLIENT=1 USE_SAVEGAMES=1 USE_FIXED_LIBAL=1 USE_SDL=1)
+    IF(USE_SYSTEM_OPENAL)
```
res2k, line 474: Would be cleaner if that bit of logic would be moved into `extern/CMakeLists.txt` as well.
Maybe add an alias target to map the "wrong" name or something.

runlevel5: @res2k I have addressed the issue by NOT setting `AL_LIBTYPE_STATIC` if `USE_SYSTEM_OPENAL` is `ON`

PR #395, opened 2024-03-27
Bump openal-soft to 1.23.1

runlevel5:
## TODO

* [ ] Addresses all warnings


apanteleev:
Getting many new warnings with the new version: `overriding '/W3' with '/W4'` for every OAL file, also implicit type conversions from `size_t` to `int` and `double` to `float`. I'd avoid updating to this version unless it brings noticeable fixes or improvements.

runlevel5:
> Getting many new warnings with the new version: overriding '/W3' with '/W4' for every OAL file, also implicit type conversions from size_t to int and double to float. I'd avoid updating to this version unless it brings noticeable fixes or improvements.

Fair call. I would prefer to keep it up to date to make things easier for downstream packager. How about I address all of the warnings first then we could revisit this PR?

apanteleev:
> How about I address all of the warnings first then we could revisit this PR?

Sounds good.

PR #403, opened 2024-04-17
Tweak USE_SYSTEM_xxx support

res2k:
The `USE_SYSTEM_xxx` logic wasn't always picking up the libs - see e.g. https://github.com/NVIDIA/Q2RTX/commit/b8fd7748249dd40fca976d18f7802f5e73f8deac#r141076760

In general, comparing the CMake module documentation with what the "package find" mechanism would provide, and what was actually linked, it seems those were actually "direct library links" instead of CMake target links.

Anyhow, I changed the CMake targets around a bit to get "wrapper" targets that pull in either the documented outputs from CMake's module detection (if "system lib" was chosen), or the "extern" library as produced "in build".
This has also the benefit that consumers of libs don't have to concern themselves with knowing which variant is taken, making src/CMakeLists.txt a bit cleaner.


runlevel5 (approved):
Much thanks for improving my works

PR #433, opened 2025-02-19
Sync with Q2PRO: Miscellaneous

res2k:
Another bunch of upstream fixes that didn't quite fit anywhere else.

### DIFF
--- a/inc/common/mdfour.h
+++ b/inc/common/mdfour.h
@@ -21,7 +21,7 @@ with this program; if not, write to the Free Software Foundation, Inc.,
 
 typedef struct mdfour {
     uint32_t A, B, C, D;
-    uint32_t count;
+    uint64_t count;
     uint8_t block[64];
 } mdfour_t;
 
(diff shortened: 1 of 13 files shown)


apanteleev:
Thanks for doing all this maintenance @res2k !

I only partially merged this PR though, it was modifying the license file - which is probably benign, but I'd rather not do that without consulting with our legal team.

Do you have any more such changes coming up? I'm hoping to actually make the 1.8 release soon, it's been delayed enough (other things took priority - CES, then RTX Kit...)

res2k:
> I only partially merged this PR though, it was modifying the license file - which is probably benign, but I'd rather not do that without consulting with our legal team.

Fair.

> Do you have any more such changes coming up? I'm hoping to actually make the 1.8 release soon, it's been delayed enough (other things took priority - CES, then RTX Kit...)

Well, yes. There's still quite a backlog in changes from Q2PRO 😅
Though not all "ready to go".
But due to the sheer amount of "sync" changes still left it probably does _not_ make sense to wait for _all_ of them, as it would probably take quite a while.

PR #440, opened 2025-02-20
Debug line drawing

res2k:
Drawing of debug lines, as found in the Q2 remaster, [Q2REPRO](https://github.com/Paril/q2pro/tree/rerelease-game), and recent Q2PRO. The code itself is based in Q2REPRO & Q2PRO, making it an amalgamation of contributions from Jonathan <jonno.5000@gmail.com>, Andrey Nazarov <skuller32@gmail.com>, and myself.

I wanted to use some debug line drawing for some toying around with the renderer; since it's in Q2PRO it would show up in a "Sync with Q2PRO" eventually, but I didn't want to wait, so I backported it from the future.

The lines themselves are drawn as line primitives on a separate texture and later composited on the RT image. I didn't want to go through RT for the lines; simplicity, for one, but I think there's also value in not "tainting" the RT image with it. To avoid wasting VRAM for an unused feature the line drawing texture is allocated lazily.

I tried to make it work with different projections apart from rectilinear - seems to generally work okay. But given this is mainly a developer utility I think a preference for rectilinear is an acceptable limitation.

A bit of a blind spot I'm having is multi-GPU support. I don't have such a setup so I can't test whether line drawing works correctly there.

There's no GL support in this PR. While it's in Q2PRO, there have been a bunch of changes to the renderer internals made before the line drawing was added. So GL support would follow some time in the future, after the prerequisite changes have arrived.

Finally, if you want to try the feature out, try playing around with the `cl_testdebuglines` cvar.


inc/refresh/debug.h
```diff
@@ -0,0 +1,42 @@
+/*
+Copyright (C) 1997-2001 Id Software, Inc.
```
apanteleev, line 2: Please use current year and your name in the copyright headers for new files.

src/refresh/debug_text.c
```diff
@@ -0,0 +1,200 @@
...
+with this program; if not, write to the Free Software Foundation, Inc.,
+51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+#include "shared/shared.h"
+#include "shared/debug.h"
+#include "common/cvar.h"
+#include "common/math.h"
+#include "common/prompt.h"
+#include "debug_fonts/cursive.h"
```
apanteleev, line 24: Do we really need all these fonts just for debug purposes? Maybe just keep one or two?

res2k: Not strictly. They're fun, though 😅

apanteleev:
> A bit of a blind spot I'm having is multi-GPU support. I don't have such a setup so I can't test whether line drawing works correctly there.

I tested this PR with a 2x RTX6000 setup, and the debug lines seem to work. There's some occasional ghosting in the scene, but I don't think it's caused by your PR. Very few people, if anyone, still use SLI setups, and it hasn't been tested in a while; both game and driver issues could've been introduced in these years.

inc/refresh/debug.h
res2k: Sure.
It's how it came from upstream and I didn't pay much mind to it.
I extended the notices with my name, and to give proper credit, the others I amalgamated the changes from.

res2k:
> I tested this PR with a 2x RTX6000 setup, and the debug lines seem to work. There's some occasional ghosting in the scene, but I don't think it's caused by your PR. Very few people, if anyone, still use SLI setups, and it hasn't been tested in a while; both game and driver issues could've been introduced in these years.

Thanks for checking!
FWIW, even though SLI is extremely rare nowadays - with Q2RTX theoretically still supporting it, I wanted to at least mention my concerns.

src/refresh/debug_text.c
apanteleev: Let's keep 1-3 fonts that are most practical and remove the rest.

res2k: Well, ok.

Issue #375, opened 2024-02-05
Native wayland small viewport & stretched graphics


FireBurn:
When building with X11 disabled or running with SDL_VIDEODRIVER=wayland the viewport appears wrong and stretched

Taking a screenshot makes things look worse 

![Screenshot_20240205_100646](https://github.com/NVIDIA/Q2RTX/assets/86515/7d7a2b08-4ae3-4350-a829-822ed53a7060)

Running with Xwayland everything works fine

This is on Q2RTX from the latest git, the latest git mesa drivers on an AMD 6800M prime laptop on Plasma 6 also from git - other native wayland games display correctly

```
SDL_VIDEODRIVER=wayland ./q2rtx
------- FS_Init -------
Current search path:
/home/fireburn/.local/share/quake2rtx/baseq2
/home/fireburn/.local/share/quake2rtx/baseq2/q2rtx_media.pkz (4054 files)
/home/fireburn/.local/share/quake2rtx/baseq2/blue_noise.pkz (258 files)
/home/fireburn/.local/share/quake2rtx/baseq2/pak2.pak (2 files)
/home/fireburn/.local/share/quake2rtx/baseq2/pak1.pak (279 files)
/home/fireburn/.local/share/quake2rtx/baseq2/pak0.pak (3307 files)
/home/fireburn/Q2RTX/baseq2
3588 files in PAK files
4312 files in PKZ files
-----------------------
Logging console to logs/console.log
Execing default.cfg
Execing q2rtx.cfg
Execing q2config.cfg

Engine version: q2rtx 1.7.0-master-5e2c4668, built on Jan 30 2024

Using SDL video driver: wayland
Loaded 2388 materials from materials/baseq2.mat
----- init_vulkan -----
Available Vulkan layers: 
  VK_LAYER_VALVE_steam_overlay_64
  VK_LAYER_VALVE_steam_overlay_32
  VK_LAYER_VALVE_steam_fossilize_64
  VK_LAYER_VALVE_steam_fossilize_32
  VK_LAYER_FROG_gamescope_wsi_x86_64
  VK_LAYER_VKBASALT_post_processing
  VK_LAYER_MESA_device_select
  VK_LAYER_MESA_overlay
  VK_LAYER_KHRONOS_validation
Vulkan instance extensions required by SDL2: 
  VK_KHR_surface
  VK_KHR_wayland_surface
Supported Vulkan instance extensions: 
  VK_KHR_device_group_creation (requested)
[...]

Issue #441, opened 2025-02-24
Crosshair mis-alignment issue


wenqingli:
CPU: AMD Ryzen 9 5950X 16-Core Processor
GPU: Radeon RX 6800 XT
System: Ubuntu 24.04.2 LTS
Desktop & display: gnome on wayland, display resolution 1920x1080
Xwayland version: The X.Org Foundation Xwayland Version 23.2.6 (12302006)

### Description
Launch `./q2rtx`, change in-game resolution from `1920x1080` to `1600x900`, then play a new game, you can observe the crosshair issue:

![Image](https://github.com/user-attachments/assets/816be639-8745-4473-8666-074d5fca3e59)

The crosshair is always in the center of the screen, but the gun is not aiming at the center of the screen in above picture, this happens when the in-game resolution is not same as the display/screen resolution.

NOTE: 
1. The issue does not occur if I switch desktop to Xorg.
2. I have checked that the call to `SDL_SetWindowDisplayMode` fails when changing the resolution, the mode is actually not changed.

res2k:
I've made some SDL2-related changes in #456, please check if they help.

wenqingli:
> I've made some SDL & Wayland related changes in [#456](https://github.com/NVIDIA/Q2RTX/pull/456), please check if they help.

Unfortunately, the issue is still there.

According to my analysis, the misalignment is caused by the projection matrix here
https://github.com/NVIDIA/Q2RTX/blob/master/src/refresh/vkpt/main.c#L2732
```c
		float viewport_proj[16] = {
			[0] = (float)fd->width / (float)qvk.extent_unscaled.width,
			[12] = (float)(fd->x * 2 + fd->width - (int)qvk.extent_unscaled.width) / (float)qvk.extent_unscaled.width,
			[5] = (float)fd->height / (float)qvk.extent_unscaled.height,
			[13] = -(float)(fd->y * 2 + fd->height - (int)qvk.extent_unscaled.height) / (float)qvk.extent_unscaled.height,
			[10] = 1.f,
			[15] = 1.f
		};
```
above matrix adds a translation effect on the standard perspective projection, so the gun no longer points to the screen center (the crosshair position).

res2k:
> According to my analysis, the misalignment is caused by the projection matrix here

Hm, maybe one or more of the _inputs_ to that matrix computation doesn't line up with the actual resolution - plausible, since you're experiencing the issue after a resolution change.
[...]

PR #456, opened 2025-03-16
Wayland-related fixes

res2k:
Came across some issues when trying to investigate #441.

Main things solved:
* VKPT renderer having incorrect belief of output resolution
* Use drawable size instead of window size (relevant in high DPI setups)

Also updated SDL2 for good measure.


FireBurn:
This fixed https://github.com/NVIDIA/Q2RTX/issues/375 for me, feel to add my

 Tested-by: Mike Lothian <mike@fireburn.co.uk>

 if they do that here

PR #465, opened 2025-04-09
Update openal-soft to 1.24.3 (again)

res2k:


apanteleev:
Thanks, but I would prefer to avoid pulling third-party actions just to update cmake. Maybe there's a way to do it with like `apt-get install...`?

Also, if the errors are indeed caused by openal-soft requiring a newer cmake verison, perhaps it's worth filing a bug against openal-soft to update their `cmake_minimum_required` which currently states `3.13`.

res2k:
> Thanks, but I would prefer to avoid pulling third-party actions just to update cmake. Maybe there's a way to do it with like `apt-get install...`?

Unfortunately, no.
The "sniper" image is based on Debian Bullseye (AFAICS). The newest CMake version there seems to be cmake 3.15.1 via bullseye-backports (https://packages.debian.org/bullseye-backports/devel/cmake) - this is the version already contained in the sniper image and just short of what we need.

> Also, if the errors are indeed caused by openal-soft requiring a newer cmake verison, perhaps it's worth filing a bug against openal-soft to update their `cmake_minimum_required` which currently states `3.13`.

It already was: https://github.com/kcat/openal-soft/issues/1088
Apparently, since the issue only affects static builds, it seems the thinking is to leave the minimum required version at 3.13.

An alternative to using a 3rd party action would be to directly pull & use the Linux x86_64 installer from https://cmake.org/download/.

res2k:
> It already was: [kcat/openal-soft#1088](https://github.com/kcat/openal-soft/issues/1088)

And also: https://github.com/kcat/openal-soft/pull/1099

apanteleev:
OpenAL updated in a464d1075225e32f2bcd036381dd12e7a95e0cad.
Closing.

PR #475, opened 2025-06-02
Implement ReSTIR sampling

abalfoort:
Code from Sirtsu55:
https://github.com/Sirtsu55/Q2RTX/commit/33d91e32f0a7bb194280e787e4ef4fc43aeb031a

I was reading through the issues and found [this one](https://github.com/NVIDIA/Q2RTX/issues/407) and remembered I saw an implementation of ReSTIR of Sirtsu55.

So, all credits go to Sirtsu55. I don't know if this is even wanted, but it was a relative small change.

PR #477, opened 2025-06-10
ReSTIR (with tweaks)

res2k:
This adds ReSTIR light sampling, based on https://github.com/zyanidelab/Q2RTX.

It's essentially #475 but some additional flourishes:
* Addressed the omissions in https://github.com/NVIDIA/Q2RTX/pull/475#issuecomment-2937189856 (although, in the meantime, use of spherical triangles has been added to #475 as well)
* Split up the original monolithic commit a bit, mainly to ease updating, but also to clearer illustrate the contained changes.

I also tried if having separate shaders for ReSTIR vs RIS makes any difference - but it didn't seem to significantly impact performance, so I kept things as they were originally.

Visuals:
* In places with noisy direct lighting, ReSTIR visible reduces the noise. (`q2dm5`, `(-605.3, -310.2, 430.2) (0.999, -0.041, -0.026)` is one such spot.)
* However, with ReSTIR enabled I observed visible dark "smears" when previously occluded pixels are revealed. This can be seen when looking at "revealed" wall pixels when the view weapon is moved, or when turning the camera. (Perhaps some unfortunate interaction with the denoiser?) Does not occur with RIS.
* Also notably, 232c47c64f07622db9ad4517bf256147aa9765b5 combines static and dynamic lights into a single list. This has the potential to degrade visuals in certain circumstances - but practically I didn't notice any issues.

Performance:
I measured frametimes using the timedemos `crusher`, `demo1`, `demo2` on a 5070 TI at 3800x2000 and a 3070 Laptop at 2560x1600. `pt_num_bounces` was set to `0`.

5070 TI:
| demo | `master` branch frametime | restir branch, `pt_restir 1` | percentage of `master` | restir branch, `pt_restir 0` | percentage of `master` |
| ----- | ----- | ----- | ----- | ----- | ----- |
| crusher | 9.551 ms | 9.593 ms | 100.4% | 9.331 ms | 97.7% |
| demo1 | 8.926 ms | 9.065 ms | 101.5% | 8.954 ms | 100.3% |
| demo2 | 9.844 ms | 9.393 ms | 95.4% | 9.328 ms | 94.8% |

3070 Laptop:
| demo | `master` branch frametime | restir branch, `pt_restir 1` | percentage of `master` | restir branch, `pt_restir 0` | percentage of `master` |
| ----- | ----- | ----- | ----- | ----- | ----- |
| crusher | 14.942 ms | 15.179 ms | 101.6% | 14.830 ms | 99.3% |
| demo1 | 14.174 ms | 14.167 ms | 99.9% | 14.179 ms | 100.0% |
| demo2 | 14.966 ms | 15.043 ms | 100.5% | 15.049 ms | 100.6% |

Since we're dealing with frametimes, lower is better.

My thoughts:
Performance-wise, there's no issue (not much of a change compared to the current light sampling), and visually, it really helps in some previously noisy areas - so it would be a clear win _if_ there wasn't the increase of "dark smearing" being observed. (Personally, I find the smearing more distracting than the noise issues, mainly because the smearing occurs constantly, while the current light sampling is decently tuned for the Q2 content and noise is not visible that often.) So it's a bit of a "so-so" right now.
But hopefully the smearing can be addressed, though I didn't look into that yet.

### DIFF
--- a/baseq2/q2rtx.menu
+++ b/baseq2/q2rtx.menu
@@ -147,6 +147,7 @@ begin video
         toggle "AMD FSR 1.0" flt_fsr_enable
         range --status "lower is sharper" "AMD FSR 1.0 sharpness" flt_fsr_sharpness 0 2 0.01
         pairs "global illumination" pt_num_bounce_rays low 0.5 medium 1 high 2
+        pairs "direct light sampling" pt_restir "RIS light" 0 "ReSTIR, high quality" 1 "ReSTIR, half" 2 "ReSTIR, quarter" 3
         pairs "reflection/refraction depth" pt_reflect_refract off 0 1 1 2 2 4 4 8 8
 		toggle --status "turn some monitors in the game into security camera views" \
             "security cameras" pt_cameras
--- a/cmake/compileShaders.cmake
+++ b/cmake/compileShaders.cmake
@@ -16,6 +16,7 @@ set(SHADER_SOURCE_DEPENDENCIES
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/precomputed_sky.glsl
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/precomputed_sky_params.h
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/projection.glsl
+    ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/restir.h
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/sky.h
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/tiny_encryption_algorithm.h
     ${CMAKE_SOURCE_DIR}/src/refresh/vkpt/shader/tone_mapping_utils.glsl
--- a/doc/client.md
+++ b/doc/client.md
@@ -937,6 +937,14 @@ Selects the projection to use for rendering. Default value is 0.
 #### `pt_reflect_refract`
 Number of reflection or refraction bounces to trace. Default value is 2.
 
+#### `pt_restir`
+Switch for experimental direct light sampling algorithms. Default value is 1.
+
+- 0 — RIS light sampling.
+- 1 — ReSTIR, high quality.
+- 2 — ReSTIR El-Cheapo, uses half of the shadow rays.
+- 3 — ReSTIR El-Very-Cheapo, uses one quarter of the shadow rays.
+
 #### `pt_roughness_override`
 Global override for roughness of all materials. Negative values mean there is no
 override. Default value is -1.
--- a/doc/client.md
+++ b/doc/client.md
@@ -939,7 +939,6 @@ Number of reflection or refraction bounces to trace. Default value is 2.
 
 #### `pt_restir`
 Switch for experimental direct light sampling algorithms. Default value is 1.
-
 - 0 — RIS light sampling.
 - 1 — ReSTIR, high quality.
 - 2 — ReSTIR El-Cheapo, uses half of the shadow rays.
--- a/inc/refresh/models.h
+++ b/inc/refresh/models.h
@@ -118,6 +118,7 @@ typedef struct light_poly_s {
 	int cluster;
 	int style;
 	float emissive_factor;
+	int type;
 } light_poly_t;
 
 typedef struct maliasmesh_s maliasmesh_t;
--- a/src/refresh/vkpt/bsp_mesh.c
+++ b/src/refresh/vkpt/bsp_mesh.c
@@ -959,7 +959,7 @@ collect_one_light_poly_entire_texture(bsp_t *bsp, mface_t *surf, mtexinfo_t *tex
 
 		light.material = texinfo->material;
 		light.style = light_style;
-		light.type = DYNLIGHT_POLYGON;
+		light.type = LIGHT_POLYGON;
 
 		if(!get_triangle_off_center(light.positions, light.off_center, NULL, 1.f))
 			continue;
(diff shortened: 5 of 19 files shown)


res2k:
> Is there a map where the smearing issue is most visible. I haven't noticed it (built and tested on Debian 12).

Hmm. It kinda appears on _every_ map. Though, curiously, certain sky-lit surfaces don't seem affected.

Here's a screengrab from `q2dm5`, near `(-605.3, -310.2, 430.2) (0.999, -0.041, -0.026)` which I mentioned earlier as a good spot to see improvements:
image

Note that this was made with `pt_num_bounces 0`. Enabling light bounces might possibly hide the issue somewhat.

abalfoort:
I see what you mean. It is even visible without setting pt_num_bounces.
I did not see that smearing in #475.

I think it's going to be difficult to track this one down.

res2k:
> I did not see that smearing in #475.

Hmm, I can notice a slight amount there. Though that might just the "unavoidable" minimum that occurs while not many samples have been accumulated yet.

Also, we have two generally working but, evidently, slightly differently behaving versions... maybe we can figure something out with the help of the differences.

abalfoort:
I haven't got much time the next couple of days. But for your convenience I've made a diff between the restir branch and this PR.
[restir_477.zip](https://github.com/user-attachments/files/20707685/restir_477.zip)

abalfoort:
My average C++ knowledge is, unfortunately, not sufficient to solve the smearing issue. I hope that @res2k finds the time to take a look at this.

apanteleev:
Thanks for working on this, @res2k and @abalfoort!

I took a quick look, and I'm generally not opposed to these changes, but they need some polishing. Most importantly, the new lighting pass needs to play well with the denoiser, and currently it doesn't. Even in the starting location of `base1`, you can see black smearing when looking around, which is not there when `pt_restir = 0`, and that means the denoiser is not getting the right signals somewhere. Looking at the signal without the denoiser (`flt_enable = 0`), you can see that the signal becomes somewhat noisier when you enable ReSTIR, which doesn't sound right.

I should be able to help you with these issues, but for that I need to allocate some time and take a closer look.

res2k:
> Even in the starting location of `base1`, you can see black smearing when looking around,

Indeed. One thing I noticed, when using Nsight Graphics, is that the "reservoir" pixels (on `PT_RESTIR_x`) in the smear area seem to be "invalid", apparently for a relatively long time, and I guess that's why don't get good lighting for these.
Unfortunately, my understanding of ReSTIR is quite limited at this time, so I don't know why that may happen. (I might also be totally wrong.)

> Looking at the signal without the denoiser (`flt_enable = 0`), you can see that the signal becomes somewhat noisier when you enable ReSTIR, which doesn't sound right.

Hmm, notable differences in light selection for ReSTIR:
* Looks at fewer lights (`RESTIR_SAMPLING_M`=4, vs `MAX_BRUTEFORCE_SAMPLING`=8)
* Doesn't consider light stats
Could those differences be relevant?

> I should be able to help you with these issues, but for that I need to allocate some time and take a closer look.

Thanks, your suggestions and input are appreciated!

abalfoort:
Just to verify if it makes any change at all I set `RESTIR_SAMPLING_M=8` and saw far less smearing. I don't know what it does to performance, but playing a bit I did not notice much change.

res2k:
Prodding around I noticed that disabling the use of the "previous reservoir" seemed to fix the smear (in exchange for more noise).
So I've pushed a commit that limits the reuse of the previous reservoir to only when it's similar in position and normal.
This seems to improve things/avoid the "smear".
Let me know what you think.

abalfoort:
Holidays are over and I've been testing the MPs with this PR.
Unfortunately, the smearing is still very much present.

res2k:
> the MPs

Uh, not sure what "the MPs" means? Could you expand the acronym?

abalfoort:
Ah, sorry, MP=Mission Pack.
I tested with MP1, The Reckoning; MP2, Ground Zero, and Zaero.

abalfoort:
Ah, well thanks for the update.

abalfoort:
@res2k How do you test the demos? I know how to use `timedemo 1` and load `demo demo1.dm2`, but not how you did your tests in the OP.

abalfoort:
Hardware info |  
-- | --
Operating System | SolydXK 13 (Debian Trixie, stable)
KDE Plasma Version | 6.3.6
KDE Frameworks Version | 6.13.0
Qt Version | 6.8.2
Kernel Version | 6.12.48+deb13-amd64 (64-bit)
Graphics Platform | Wayland
Processors | 16 × AMD Ryzen 7 3700X 8-Core Processor
Memory | 64 GiB of RAM (62,7 GiB usable)
Graphics Processor | NVIDIA GeForce RTX 3060
Manufacturer | Gigabyte Technology Co., Ltd.
Product Name | B550 AORUS ELITE V2
Nvidia driver | 580.105.08
Resolution | 2560x1440 @179,96, full screen


demo | master branch frametime | restir-update branch, pt_restir 0, ms | percentage of master | restir-update branch, pt_restir 1, ms | percentage of master
-- | -- | -- | -- | -- | --
crusher | 21,253 | 21,216 | 100,2 | 21,253 | 100,0
demo1 | 20,844 | 19,908 | 104,7 | 20,247 | 102,9
demo2 | 22,127 | 21,472 | 103,0 | 21,732 | 101,8

abalfoort:
It is map city1 (Outer Courts).
How do I disable indirect lighting?

res2k:
> I guess you do this by setting

`pt_num_bounce_rays 0`

apanteleev:
The project is discontinued; no new features will be implemented here. Closing.

PR #481, opened 2025-06-20
Add permadeath and ironman mode with savegame support and ammo/health balancing

kendrixargon:
This feature introduces a new permadeath and ironman modes with the following behavior:
- Uses CVAR_LATCH to protect against in-game modification;
- Integrates with the main menu as an extra difficulties;
- State is preserved through save/load (saved in gamefields);
- You can only quicksave in this mode, normal saves are blocked;
- When you die game deletes quicksave and autosave.

Ironman difficulty has additional balance rules:
- Ammo pickups reduced by approx. 50%;
- Medkits & large medkits heals for 80%.

No additional CVARs are required, and behavior is consistent with other latched variables (e.g. deathmatch, coop, skill).


kendrixargon:
I understand your concerns, had doubts myself, it's because most of you folks treat this as technical playground, but i believe there's another group of people like me that actually enjoy playing game in this version. Subjectivly this is successor to anything that already exists.

Consider that my solution as more accessible/effortless because you don't need to find, download, and load mod into source port manually, this can be obstacle for less technical or lazy users. It's more like "plug and play" approach.

My motivation to make this was to challange myself, because i find Quake 2 as one of the easiest shooters. Finishing this about 30-40 times doesn't help, believe me ;)

In the matter of standalone mod that could be next step for my sparetime project, but i have doubts that i can reimplement this 1:1 as a mod. I've extended the ui which gonna need different approch for mod(if possible), based on my small research.

Unfortunatlly cannot move permadeath save block to WriteGame() because no matter how i save the game filename always contains "./baseq2/save/.current/game.ssv"

Already finished game on this difficulty, I'll commit some polishing changes along with new variable "rebalance" to separate it from "permadeath".
https://www.youtube.com/watch?v=-a-kAbAsGjM

kendrixargon:
Alright, thank your for your opinions, im closing this pull request, gonna share this by other means.

